PR #1, opened 2018-02-07
Problem: there is no collaboration model

vrde:
Solution: use the ZeroMQ C4 collaboration model.

We can use this PR to fine tune C4 to our needs. Please note that the status of this proposal is **raw**. We should keep it **raw** or **draft** until we are comfortable with it.

### DIFF
--- a/1/README.md
+++ b/1/README.md
@@ -42,7 +42,7 @@ C4 is meant to provide a reusable optimal collaboration model for open source so
 
 ### 2.2. Licensing and Ownership
 
-1. The project SHALL use a license such as Apache 2, MIT, MPLv2, or a GPLv3 variant thereof (GPL, LGPL, AGPL).
+1. The project SHALL use the Apache 2 license.
 1. All contributions to the project source code ("patches") SHALL use the same license as the project.
 1. A new contributor contributing a patch MUST agree to a Contributor License Agreement (CLA) before their patch can be merged.
 1. Each Contributor SHALL be responsible for identifying themselves in the project Contributor list.
--- a/1/README.md
+++ b/1/README.md
@@ -42,7 +42,7 @@ C4 is meant to provide a reusable optimal collaboration model for open source so
 
 1. The project SHALL use a license such as Apache 2, MIT, MPLv2, or a GPLv3 variant thereof (GPL, LGPL, AGPL).
 1. All contributions to the project source code ("patches") SHALL use the same license as the project.
-1. There MAY be a copyright assignment process for patches submitted by new Contributors.
+1. A new contributor contributing a patch MUST agree to a Contributor License Agreement (CLA) before their patch can be merged.
 1. Each Contributor SHALL be responsible for identifying themselves in the project Contributor list.
 
 ### 2.3. Patch Requirements
--- a/1/README.md
+++ b/1/README.md
@@ -1,0 +1,121 @@
+```
+shortname: 1/C4
+name: Collective Code Construction Contract
+status: raw
+editor: Alberto Granzotto <alberto@bigchaindb.com>
+```
+
+## Abstract
+
+C4 provides a standard process for contributing, evaluating and discussing improvements on software projects. It defines specific technical requirements for projects like a style guide, unit tests, `git` and similar platforms. It also establishes different personas for projects, with clear and distinct duties. C4 specifies a process for documenting and discussing issues including seeking consensus and clear descriptions, use of "pull requests" and systematic reviews.
+
+## Language
+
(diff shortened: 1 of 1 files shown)


codegeschrei:
2.4.10
do we need this rule? Do we want PRs as issues? 
I'd prefer to have everything separated. I think having both can lead to a lack of structure. Issues and PRs have their own context and description. Though if you have both they have to be referenced, otherwise it's leads to a search for the matching issue/PR.
In a recent discussion @sbellem pointed out that the missing structure can lead to problems with bigger projects. Could you maybe give some input on this here @sbellem 

2.6.2
In a discussion we talked about the `space` in this rule and if it should be a public space or not. If we want it to be a public space I would add it to the rule.

1/README.md
```diff
@@ -40,9 +40,9 @@ C4 is meant to provide a reusable optimal collaboration model for open source so
 
 ### 2.2. Licensing and Ownership
 
-1. The project SHALL use a share-alike license such as the MPLv2, or a GPLv3 variant thereof (GPL, LGPL, AGPL).
+1. The project SHALL use a license such as Apache 2, MIT, MPLv2, or a GPLv3 variant thereof (GPL, LGPL, AGPL).
```
ttmc: I guess these changes make this document the BigchainDB variant of the C4? Maybe we could indicate that in the title, just so it's clear that it's not the same as the original/standard C4? There could even be a "diff document" listing the differences between the two.

As for the license, I think Bruce's current thinking is that all BigchainDB software will be licensed Apache v2 for now. Not other licenses.

1/README.md
```diff
@@ -40,9 +40,9 @@ C4 is meant to provide a reusable optimal collaboration model for open source so
...
 1. All contributions to the project source code ("patches") SHALL use the same license as the project.
-1. All patches are owned by their authors. There SHALL NOT be any copyright assignment process.
+1. There MAY be a copyright assignment process for patches submitted by new Contributors.
```
ttmc: The BigchainDB CLA isn't a copyright assignment. People who sign that CLA keep their copyright and give BigchainDB GmbH the right to license their contribution under an open source license (such as Apache v2). Maybe:

`1. A new contributor contributing a patch MUST agree to a Contributor License Agreement (CLA) before their patch can be merged.`

1/README.md
vrde: > I guess these changes make this document the BigchainDB variant of the C4? Maybe we could indicate that in the title, just so it's clear that it's not the same as the original/standard C4? There could even be a "diff document" listing the differences between the two.

Yes, this is our variant. [unprotocols.org 1/C4](https://github.com/unprotocols/rfc/blob/master/1/README.md) says _This RFC is equivalent (with the exception of minor cosmetic changes) to ZeroMQ RFC 42/C4_. I can say that this C4 is a customized version of the one by unprotocols. About the diff, users can do it themselves, it's really easy and we have one less document to maintain.

> As for the license, I think Bruce's current thinking is that all BigchainDB software will be licensed Apache v2 for now. Not other licenses.

I was keeping it open in case we start other projects with different requirements.

ttmc: We can always modify this document in the future, if we decide to license some projects with a license other than Apache v2.

ttmc:
I mentioned this already, but maybe I could put it another way...

If someone comes to this page, they might already be familiar with C4 and think, "Oh, they're using C4! I know C4." There's no up-front indication that we've modified the standard C4, or how. Yes, they _could_ do a diff, but will they? I suspect most won't.

vrde:
> 2.4.10
> If the Platform implements pull requests as issues, a Contributor MAY directly send a pull request without logging a separate issue.
> do we need this rule? Do we want PRs as issues?
> I'd prefer to have everything separated. I think having both can lead to a lack of structure. Issues and PRs have their own context and description. Though if you have both they have to be referenced, otherwise it's leads to a search for the matching issue/PR.
> In a recent discussion @sbellem pointed out that the missing structure can lead to problems with bigger projects. Could you maybe give some input on this here @sbellem

What kind of rule would you enforce?

Writing an issue first makes sure that the problem you are identifying is an actual problem. When your issue is acknowledge by the community, then you should write code. But if you are really sure what you are doing is valuable, you can always "risk" and you immediately file a PR. Worst case you lost time writing code :)

vrde:
Should we keep the list of maintainers in `MAINTAINERS` under the branch `master`? C4 doesn't specify this.

codegeschrei:
> What kind of rule would you enforce?
>Writing an issue first makes sure that the problem you are identifying is an actual problem. When your issue is acknowledge by the community, then you should write code. But if you are really sure what you are doing is valuable, you can always "risk" and you immediately file a PR. Worst case you lost time writing code :)

So when it is an acknowledge issue by the community, maybe the issue could just link to the place where people acknowledged the problem or it is summarised in the issue?

> Should we keep the list of maintainers in MAINTAINERS under the branch master? C4 doesn't specify this.

I think this is a good idea.

vrde:
> I think this is a good idea.

I changed my mind. `MAINTAINERS` is another thing to keep up to date. Let's merge as it is, this specification is still `raw`.

PR #2, opened 2018-02-08
Problem: there is no way to propose improvements

vrde:
Solution: have a set of guidelines to handle that. Instead of doing it from scratch, an idea is to use COSS.

This is a work in progress. I'm making this PR to discuss how to adapt COSS to our needs.


2/README.md
```diff
@@ -0,0 +1,122 @@
...
+* The process should be visible, objective, and accessible to anyone.
+* The process should clearly separate experiments from solutions.
+* The process should allow deprecation of old specifications.
+
+Specifications should take minutes to explain, hours to design, days to write, weeks to prove, months to become mature, and years to replace.
+
+Specifications have no special status except that accorded by the community.
+
+## Architecture
+COSS is designed around fast, easy to use communications tools. Primarily, COSS uses a wiki model for editing and publishing specifications texts.
```
vrde: I'm confused by *wiki model*. Is keeping proposals under a GitHub repo a *wiki model*?

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+* The *domain* is the conservancy for a set of specifications in a certain area.
```
vrde: IMO we can drop the need to specify the *domain*. I don't see a big advantage specifying it. The domain is the one of the organization where the `rfc` repository is.

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+* Each domain is implemented as an Internet domain, hosting a wiki and optionally other communications tools.
```
vrde: Both the ZeroMQ community and the Unprotocols community have dedicated subdomains, specifically:
- https://rfc.zeromq.org/
- https://rfc.unprotocols.org/

I like this, but it's *one more thing™* to care about. What if we keep everything in the GitHub repository? [Bitcoin](https://github.com/bitcoin/bips) and [Ethereum](https://github.com/ethereum/EIPs) do that.

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+* Each specification is a set of wiki pages, together with comments, attached files, and other resources.
```
vrde: Instead of `wiki pages`, I'd say `markdown documents`. I don't understand what `together with comments` means.

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+* Important specifications may also exist as subdomains, i.e. child wikis.
```
vrde: To me, this does not make a lot of sense.

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+Individuals can become members of the domain by completing the necessary legal clearance. The copyright, patent, and trademark policies of the domain must be clarified in an Intellectual Property policy that applies to the domain.
```
vrde: Is it related to the intellectual property of the proposal?
If so, [PEP says](https://www.python.org/dev/peps/pep-0001/#what-belongs-in-a-successful-pep):
> Copyright/public domain -- Each PEP must either be explicitly labeled as placed in the public domain (see this PEP as an example) or licensed under the Open Publication License [7].

[BIP says](https://github.com/bitcoin/bips/blob/master/bip-0002.mediawiki#BIP_licensing):
> New BIPs may be accepted with the following licenses. Each new BIP must identify at least one acceptable license in its preamble. The License header in the preamble must be placed after the Created header. Each license must be referenced by their respective abbreviation given below.

What is fun:
- I don't find anything related to licensing for the Ethereum Improvement Proposal guidelines
- Python Enhancement Proposal guidelines suggest to use the Open Documentation License, while Bitcoin Improvement Proposal guidelines say that Open Documentation License is **not** an acceptable license.

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+Specifications exist as multiple pages, one page per version of the specification (see "Branching and Merging", below), which may be assigned URIs that include an incremental number. Thus, we refer to a specification by specifying its domain, number, and short name. New versions of the same specification will have new numbers. The syntax for a specification reference is:
+
+
+  <domain>/spec:<number>/<shortname>
+
+
+For example, this specification is **rfc.unprotocols.org/spec:2/COSS**. The short form **2/COSS** may be used when referring to the specification from other specifications in the same domain.
```
vrde: I'd simplify all of this saying that every specification lives in a directory under the GitHub repository.

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+Every specification (including branches) carries a different number. Lower numbers indicate more mature specifications, higher numbers indicate more experimental specifications.
+
+## COSS Lifecycle
+Every specification has an independent lifecycle that documents clearly its current status.
+
+A specification has six possible states that reflect its maturity and contractual weight:
+
+TBD: Add lifecycle
```
vrde: I need to add the image here.

2/README.md
codegeschrei: I was also confused. My interpretation: maybe they mean the [gitbook](https://rfc.unprotocols.org/spec:2/COSS/)

2/README.md
codegeschrei: I think we can start of having it in GitHub and if we want we can still move it to a subdomain later on

2/README.md
kansi: I guess that would be the `c` part of `rfc`?

2/README.md
kansi: Maybe the directory naming conventions should also be defined? we could adapt the above convention in a directory naming scheme?

2/README.md
vrde: Oh well, yes 😸

Every specification starts with a PR, so `together with comments` can refer to those comments.

2/README.md
vrde: I agree, my idea is to have something simple, like
- `bigchaindb/rfc` this repository
  - `README.md` with pointers to C4 and COSS
  - `1/` directory for RFC 1
    - `README.md` the actual specification
    - `...` all other media files needed.
  - `2/` directory for RFC 2
    - `README.md` the actual specification
    - `...` all other media files needed.

2/README.md
ttmc: Maybe they mean a model where "anyone can edit," like in a wiki?

2/README.md
ttmc: Maybe replace *domain* with *Git repository*. I don't know if we need to tie it to a specific URL. In theory, one can migrate a Git repository from GitHub to GitLab to Bitbucket to whatever.

2/README.md
codegeschrei: I'd say we remove it. If we ever see the need of a subdomain we can add it again?

2/README.md
ttmc: Should we specify that all the files for a particular specification should live inside *one* directory?

2/README.md
ttmc: Can we do a bulk search for "domain" and replace all?

ttmc: EIPs must be public domain. It's the last, required section of each EIP: copyright assignment. A handy way to do that is to use a CC0 Public Domain Dedication - see https://creativecommons.org/publicdomain/zero/1.0/

2/README.md
ttmc: Is the idea to name each directory 1, 2, 3, etc. --- one per specification?

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+Specifications exist as multiple pages, one page per version of the specification (see "Branching and Merging", below), which may be assigned URIs that include an incremental number. Thus, we refer to a specification by specifying its domain, number, and short name. New versions of the same specification will have new numbers. The syntax for a specification reference is:
+
+
+  <domain>/spec:<number>/<shortname>
+
+
+For example, this specification is **rfc.unprotocols.org/spec:2/COSS**. The short form **2/COSS** may be used when referring to the specification from other specifications in the same domain.
+
+Every specification (including branches) carries a different number. Lower numbers indicate more mature specifications, higher numbers indicate more experimental specifications.
```
ttmc: Hmm, I'm not sure about this. I think the numbers are just there to indicate the order that they were added to the set. Some might never get implemented. Others might get implemented fast, others slowly. The numbering therefore won't be a great way to tell how mature one is.

2/README.md
ttmc: In addition to the image, it would be handy to have a list of the six states in text form.

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+A specification has six possible states that reflect its maturity and contractual weight:
+
+TBD: Add lifecycle
+
+### Raw Specifications
+All new specifications are **raw** specifications. Changes to raw specifications can be unilateral and arbitrary. Those seeking to implement a raw specification should ask for it to be made a draft specification. Raw specifications have no contractual weight.
+
+### Draft Specifications
+When raw specifications can be demonstrated, they become **draft** specifications. Changes to draft specifications should be done in consultation with users. Draft specifications are contracts between the editors and implementers.
```
ttmc: I don't understand how one would know when a raw specification "can be demonstrated." Demonstrated? Do they mean *implemented*?

2/README.md
codegeschrei: This is a good idea but maybe it is still too specific. Some RFCs might not only be related to one repository (change of the tx model leads to changes in various places). So we could do it as a `component`. For now we could have two components `transaction model` and `http API`.
New components can be added over time. With this you can propose a change and the change needs to be applied for every repository it concerns.

2/README.md
vrde: In my opinion yes, one per specification. My suggestion is to name the main specification `README.md`, so GitHub shows it right away.

2/README.md
codegeschrei: `Each specification is a directory, together with comments, attached files, and other resources.` ?

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+## Branching and Merging
+Any member of the domain MAY branch a specification at any point. This is done by copying the existing text, and creating a new specification with the same name and content, but a new number. The ability to branch a specification is necessary in these circumstances:
+
+* To change the responsible editor for a specification, with or without the cooperation of the current responsible editor.
+* To rejuvenate a specification that is stable but needs functional changes. This is the proper way to make a new version of a specification that is in stable or deprecated status.
+* To resolve disputes between different technical opinions.
+
+The responsible editor of a branched specification is the person who makes the branch.
+
+Branches, including added contributions, are derived works and thus licensed under the same terms as the original specification. This means that contributors are guaranteed the right to merge changes made in branches back into their original specifications.
```
ttmc: If we decide to ask all contributors to dedicate all specs to the public domain (e.g. using CC0), then there is no license. This paragraph would have to be reworded, e.g.

Branches, including added contributions, SHOULD be dedicated to the public domain using CC0 (just like the original specification). This means that contributors are guaranteed the right to merge changes made in branches back into their original specifications.

2/README.md
```diff
@@ -0,0 +1,122 @@
...
+COSS resolves natural conflicts between teams and vendors by allowing anyone to define a new specification. There is no editorial control process except that practised by the editor of a new specification. The administrators of a domain (moderators) may choose to interfere in editorial conflicts, and may suspend or ban individuals for behaviour they consider inappropriate.
+
+## Conventions
+Where possible editors and contributors are encouraged to:
+
+* Refer to and build on existing work when possible, especially IETF specifications.
+* Contribute to existing specifications rather than reinvent their own.
+* Use collaborative branching and merging as a tool for experimentation.
+
+## License
```
ttmc: The GNU General Public License is for software and is completely inappropriate for something like specs, but I guess they used it because it was available at the time. Today, there are many more appropriate licenses, such as the GNU Free Documentation License (GFDL) and the various CC licenses.

As mentioned above, we could just ask contributors to dedicate their contributions (specs) to the public domain using CC0, i.e. https://creativecommons.org/publicdomain/zero/1.0/

2/README.md
vrde: Ethereum has [types](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md#eip-types). It's a really good approach in my opinion.

2/README.md
codegeschrei: I would say it's about an implementation. The next in line would be `stable` which is about the use by others, so it has to work.

2/README.md
codegeschrei: What types would be a good thing to start? `core`, `network`, ?

ttmc: I think the whole point of the /rfc repository is to have specifications that span the whole BigchainDB community / ecosystem, including server, drivers, various libraries, etc. When I wrote *Git repository* above, I meant the /rfc repository containing all the specs.

codegeschrei: I see. So since we just offer one repository we can just skip this and not mention it.

muawiakh: Components or Types or Project, all sound good to me.

2/README.md
muawiakh: +1 to removing this, they all should be treated the same, maybe have labels/tags for such scenarios.

muawiakh:
+1, for consolidating all the information and workflow here. Really good to see.

IMO, instead of PR having a `Problem: <details>` in the summary, just refer to a ticket/Issue. GitHub Issue(s) <--> Problem(s), same thing. The solution should also be part of the ticket/issue, if you know already. The PR should only describe the changes you made to accomplish it. Comments?

codegeschrei:
For now we have merged C4, which currently allows to have PRs without an issue (therefore we need a solution for both cases).
The basic `Problem: ... Solution:...` structure is for commit messages and stated in C4.
Thus this seems to be a C4 related topic.?

2/README.md
vrde: I just noticed that an EIP takes the number of the PR. We can do the same.

2/README.md
```diff
@@ -34,25 +34,26 @@ Specifications should take minutes to explain, hours to design, days to write, w
...
 
+Every specification (including branches) carries a different number. New versions of the same specification have new numbers.
 
-For example, this specification is **rfc.unprotocols.org/spec:2/COSS**. The short form **2/COSS** may be used when referring to the specification from other specifications in the same domain.
+### Specification template
+TBD: Embed Muawia work here https://github.com/bigchaindb/bigchaindb/tree/master/proposals
 
-Every specification (including branches) carries a different number. Lower numbers indicate more mature specifications, higher numbers indicate more experimental specifications.
+### Copyright Waiver
+All Specifications MUST be in public domain. The only two exceptions are 1/C4 and 2/COSS, licensed under GPL.
```
ttmc: How about, "Except for 1/C4 and 2/COSS, all Specifications MUST be released to the public domain."

(Anyone curious about how 1/C4 and 2/COSS are licensed can look in the **License** section of those documents.)

Maybe we could provide an example sentence that they could use to release the spec to the public domain. Here's one I got by using the [Creative Commons CC0 wizard](https://creativecommons.org/choose/zero/):

To the extent possible under law, the Editor and Contributors listed above have waived all copyright and related or neighboring rights to TITLE OF THIS DOCUMENT. This work is published from: INSERT COUNTRY HERE.

vrde: Thanks @ttmc, I didn't know about the CC0 wizard. If you do it without filling the form, you get:
>  To the extent possible under law, the person who associated CC0 with this work has waived all copyright and related or neighboring rights to this work.

2/README.md
vrde: I'd love to have C4 and COSS under CC0, but the original work has been licensed using GPL, so we need to release our changes under the same license.

ttmc:
There's still Muawia's template to embed or reference somehow. Maybe that could be done in a new pull request, because small pull requests are better than big ones.

PR #3, opened 2018-02-09
Spec to dynamically add/update/remove validators at runtime

kansi:
The spec discuss the challenges and implementation details for dynamically adding validators.

### DIFF
--- a/3/README.md
+++ b/3/README.md
@@ -1,4 +1,11 @@
-# Dynamically add/remove validators at runtime (Insecure version)
+```
+shortname: 3/UPSERT-VALIDATORS
+name: Dynamically add/remove validators at runtime (Insecure version)
+type: Feature
+status: raw
+editor: Vanshdeep Singh <vanshdeep@bigchaindb.com>
+```
+
 
 ## Problem Description
 In the current state of tendermint bdb, validator nodes are specified in the Tendermint config file before starting the node. It is necessary that the users should be able to dynamically add new validators.
--- a/dynamic_validators/insecure_dynamic_validators.md
+++ b/dynamic_validators/insecure_dynamic_validators.md
@@ -18,48 +18,44 @@ Note: `"power": 0` implies that the validator should be removed from the validat
 
 
 ## Proposed Change
-- `PUT /api/v1/validators` should be introduced in order to allow bdb client applications to add/remove validators. The body of the request should contain the validators diff which would then be returned during `end_block`.
+- Add a new subcommand `bighciandb upsert-validator PUBLIC_KEY POWER` which will allow the node admin to add/update/delete a validator.
 
 - `GET /api/v1/validators` should be introduced to list the current validators.
 
 ### Technical challenges
-Since bdb clients can call `PUT /api/v1/validators` anytime it is necessary to (temporarily) store the payload until `end_block` is called. This implies that a successful `PUT` doesn't necessarily mean that the validators list has been updated. Moreover, one also needs to consider a scenario wherein the Tendermint node goes offline for whatever reason. When the node is brought back online it would load the validators list from its config file which may not include dynamically added validators.
+Since node admin can call `upsert-validator` anytime it is necessary to (temporarily) store the payload until `end_block` is called. This implies that just executing `bigchaindb upsert-validator` doesn't imply the validator is add to validators list.
 
 Furthermore, as discussed in [technical details](#technical-details) each/majority (>2/3) nodes in the network need to add the new node to their list of validators in order for the new node to be able to act as a validator. Until then the node may not be able to participate and vote on blocks.
 
 
 ### Storing dynamically added validators
-As discussed in the previous section it is necessary to (temporarily) store the updates to validators list until `end_block` is executed. Furthermore, in case of a node crash it is required that that upon restart the node should be able to recover its dynamically added validators. Following is one possible solution,
+In the previous section it was discussed that it is necessary to (temporarily) store the updates to validators list until `end_block` is executed. Following is one possible solution,
 
-- Store the validators in a MongoDB collection along with a flag when `PUT /api/v1/validators` (discussed below) is called i.e.,
+- Store validator diff in a MongoDB collection along with a flag when `bighciandb upsert-validator` (discussed below) is called i.e.,
 ```json
-{"validators": [],
+{"validators": [{"pub_key":{"type":"ed25519","data":"4E2685D9016126864733225BE00F005515200727FBAB1312FC78C8B76831255A"},
+                 "power":10}],
  "sync": true
 }
 ```
-- The sync flag is used during `end_block` to decide if a validator update should be returned.
-- The sync flag is set to `true` when api `PUT /api/v1/validators` is called and the `"validators"` are updated.
-- Whenever the sync flag is set to `true`, `end_block` will call `http://tendermint:46657/validators`, calculate the diff with the list of validators retrieved from mongodb and return it.
-- When bigchaindb starts the sync flag should be set to `true`.
-- In case the sync flag is set to `false`, it implies that the validators are synced and `end_block` shall return `[]` (empty validator diff).
+- The `"sync"` key is used during `end_block` to decide if a validator update should be returned.
+- During `end_block` all the documents with `"sync": true` are aggregated in order based on their insert time stamp and validator diff list is prepared which is returned in `end_blocks` response.
 
+NOTE: a validators diff which results in a change of `>1/3` of voting power will be rejected by tendermint.
 
 ### API impact
-Following new api endpoints should be introducted,
 
-- Add/remove validators
-  - Uri: `PUT /api/v1/validators`
-  - Body: 
-  ```json
-  [{"pub_key":{"type":"ed25519","data":"4E2685D9016126864733225BE00F005515200727FBAB1312FC78C8B76831255A"},
-    "power":10},
-   {"pub_key":{"type":"ed25519","data":"608D839D7100466D6BA6BE79C320F8B81DE93CFAA58CF9768CF921C6371F2553"},
-   "power":0}
-   ]
-   ```
-   - response:
-     - `200` i.e. the request was correctly formed and the MongoDB collection has been updated
-     - `400` i.e. there was issue with the request payload i.e. value for `"pub_key"` or `"power"` was invalid.
+- A new subcommand `upsert-validator` should be introduced to add/update/delete validators
+  - syntax : `bichaindb upsert-validator PUBLIC_KEY POWER`
+  - example usage
+    - add/update validator
+    ```bash
+    $ bichaindb upsert-validator 4E2685D9016126864733225BE00F005515200727FBAB1312FC78C8B76831255A 10
+    ```
+    - delete validator
+    ```bash
+    $ bichaindb upsert-validator 4E2685D9016126864733225BE00F005515200727FBAB1312FC78C8B76831255A 0
+    ```
 
 - List validators
   - Uri: `GET /api/v1/validators`
@@ -75,30 +71,20 @@ Following new api endpoints should be introducted,
 **NOTE**: when `power: 0` is set for a given validator it is removed from the validators list.
 
 ### Workflow overview
-Below is a summary of worflow sequence which should be executed in order to add/remove a validator dynamically,
+Below is a summary of workflow sequence which should be executed in order to add/remove a validator dynamically,
 - BigchainDB client wishes to add/remove validators from their own node.
-- The client calls `PUT /api/v1/validators` with the following payload,
-```json
-[{"pub_key":{"type":"ed25519","data":"4E2685D9016126864733225BE00F005515200727FBAB1312FC78C8B76831255A"},
-  "power":10},
- {"pub_key":{"type":"ed25519","data":"608D839D7100466D6BA6BE79C320F8B81DE93CFAA58CF9768CF921C6371F2553"},
- "power":0}
- ]
-```
-where `"power": 0` indicates that the key will removed from the given node.
+- The admin use `bigchaindb upsert-validator` to add the validator,
 - BigchainDB stores the payload in a MongoDB collection as follows,
 ```json
 {"validators": [{"pub_key":{"type":"ed25519","data":"4E2685D9016126864733225BE00F005515200727FBAB1312FC78C8B76831255A"},
-                 "power":10},
-                {"pub_key":{"type":"ed25519","data":"608D839D7100466D6BA6BE79C320F8B81DE93CFAA58CF9768CF921C6371F2553"},
-                 "power":0}
+                 "power":10}
                ],
  "sync": true
 }
 ```
-- The bigchainDB client has successfully submitted a request to update the validators.
-- Now, when `end_block` is executed, the validators from MongoDB collection are aggregated where `"sync": false`. **NOTE**: the aggregation should be performed by sorting the documents using timestamp.
-- The aggregated validators are returned during `end_block` and the corresponding documents are updated i.e. `"sync": true`.
+- The request has been successfully submitted to update the validators.
+- Now, when `end_block` is executed, the validators from MongoDB collection are aggregated where `"sync": true`. **NOTE**: the aggregation should be performed by sorting the documents using time stamp.
+- The aggregated validators are returned during `end_block` and the corresponding documents are updated i.e. `"sync": false`.
 
 
 ### Security impact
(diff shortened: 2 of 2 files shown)


dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,103 @@
...
+- Change the power of a validator
+**NOTE**: The above tests should check if the MongoDB collection is updated and calling `end_block` returns a valid diff. More accurate tests could be written in the integration test suite wherein `http://tendermint:46657/validators` is called to verify the validator changes.
+
+## Implementation
+
+### Assignee(s)
+Primary assignee(s): @kansi
+
+### Targeted Release
+BigchainDB 2.0
```
ttmc: Earlier, you used the terminology "bigchaindb tendermint MVP." That's probably better, because I'm not sure what the final name or version number will be.

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+## Dependencies
+N/A
+
+
+## Reference(s)
+N/A
```
ttmc: You could reference the Tendermint docs about `end_block` and how to add/remove validators after the blockchain gets started.

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,103 @@
...
+    "power":10},
+   {"pub_key":{"type":"ed25519","data":"608D839D7100466D6BA6BE79C320F8B81DE93CFAA58CF9768CF921C6371F2553"},
+   "power":0}
+   ]
+  ```
+
+**NOTE**: when `power: 0` is set for a given validator it is removed from the validators list.
+
+### Security impact
+The current implementation is insecure i.e. any user who gains access to a given node can modify its validators. This node could effectively be considered byzantine and the network can tolerate (<1/3) byzantine nodes.
```
ttmc: It's not clear to me why this is insecure for the overall system. Here's why:

Suppose a BigchainDB node operator uses the `end_block` response to tell his local Tendermint instance to update its local list of validators. It will do that, but only locally, not on any other Tendermint instance. The "damage" is limited to just that node.

If enough node operators add the same new validator to their list (i.e. with more than 2/3 of the total power), then the new validator will be recognized as a legitimate validator and its votes will be accepted, otherwise its votes will be ignored. Is that right? If so, that seems fairly secure to me.

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Store the validators in a MongoDB collection along with a flag i.e.,
+```json
+{"validators": [],
+ "sync": true
+}
+```
+- The sync flag is used during `end_block` to decide if a validator update should be returned.
+- The sync flag is set to `True` when api `PUT /api/v1/validators` is called and the `"validators"` are updated.
+- Whenever the sync flag is set, `end_block` will call `http://tendermint:46657/validators`, calculate the diff with the list of validators retrieved from mongodb and return it.
```
ttmc: Whenever the sync flag is set to what?

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,103 @@
...
+- When bigchaindb starts the sync flag should be set to `True`.
+
```
ttmc: Maybe add a bullet point to explain what `"sync": false` would mean.

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,103 @@
...
+  - Uri: `PUT /api/v1/validators`
+  - Body: 
+  ```json
+  [{"pub_key":{"type":"ed25519","data":"4E2685D9016126864733225BE00F005515200727FBAB1312FC78C8B76831255A"},
+    "power":10},
+   {"pub_key":{"type":"ed25519","data":"608D839D7100466D6BA6BE79C320F8B81DE93CFAA58CF9768CF921C6371F2553"},
+   "power":0}
+   ]
+   ```
+   - response: `200`
```
ttmc: I guess there would be another possible HTTP response if the body was badly-formed or wrong in some way.

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,103 @@
+# Dynamic Validators (Insecure version)
```
ttmc: The title of this PR and this spec is misleading. It suggests that there are "dynamic validators" which are somehow different from the validators that existed before. I think what you mean is something like:

Spec for adding/removing validators dynamically at runtime

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,103 @@
...
+
+Furthermore, as discussed in [technical details](#technical-details) each/majority (>2/3) nodes in the network need to add the new node to their list of validators in order for the new node to be able to act as a validator. Until then the node may not be able to participate and vote on blocks.
+
+
+### Storing dynamically added validators
+As discussed in the previous section it is necessary to (temporarily) store the updates to validators list until `end_block` is executed. Furthermore, in case of a node crash it is required that that upon restart the node should be able to recover its dynamically added validators. Following is one possible solution,
+
+- Store the validators in a MongoDB collection along with a flag i.e.,
+```json
+{"validators": [],
```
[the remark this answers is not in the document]
kansi: @ttmc the `"validators"` key stores the diff. So an empty diff is valid, it would imply that the validators should not be changed.

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,142 @@
+# Dynamically add/remove validators at runtime (Insecure version)
+
+## Problem Description
+In the current state of tendermint bdb, validator nodes are specified in the Tendermint config file before starting the node. It is necessary that the users should be able to dynamically add new validators.
+
+**NOTE**: This document presents an insecure implementation intended for bigchaindb tendermint MVP.
+
+### Technical details
+Tendermint allows the abci client application to return a list of validators during `end_block` call. If the client doesn't wish to modify the validators an empty list should be returned. The return value allows the client to dynamically add/remove validators. Note that the return value could be thought of as a (validators) diff of desired validators list and current validators list. Below is an example of validators diff,
```
ttmc: Is BigchainDB supposed to return a diff / delta / change-set or a _complete_ list of validators (in the End Block response)?

After reading and re-reading [the Tendermint docs about End Block](https://tendermint.readthedocs.io/en/master/app-development.html#endblock), I now think the answer is:

- If you don't want to change the validators (or their powers) at all, then return an empty list `[]` in the End Block response.
- If you _do_ want to change the validators in some way, then _return a complete list of all validators_ with any new ones added, and with the ones you want to delete having their power set to 0.

ttmc:
Maybe we can make a secure version with the same amount of effort as making an insecure one.

How? Just have the node operator use the MongoDB API (as local MongoDB admin user) to update the new collection (the local list of validators). No new HTTP PUT endpoint would be needed. The GET endpoint might still be useful.

Why would that be secure? Suppose a BigchainDB node operator arranges for the `end_block` response to tell his local Tendermint instance to update its local list of validators (e.g. by changing the new validators-list collection). It will do that, but only locally, not on any other Tendermint instance. The "damage" is limited to just that node.

If enough node operators add the same new validator to their list (i.e. with more than 2/3 of the total power), then the new validator will be recognized as a legitimate validator and its votes will be accepted, otherwise its votes will be ignored. Is that right? If so, that seems fairly secure to me.

There could be some convenience CLI commands like `bigchaindb add-validator PUBLIC_KEY POWER` and `bigchaindb change-validator-power PUBLIC_KEY POWER` (for use by the node operator).

dynamic_validators/insecure_dynamic_validators.md
kansi: @ttmc I don't think one needs to _return a complete list of all validators_ . The documentation says,
> To add a new validator or update an existing one, simply include them in the list returned

The way I interpret this is that in case a user doesn't wish to update an existing validator they shouldn't need to include them in the list. That is whey I refered to it has a diff. 

Am I missing something?

ttmc: > To add a new validator or update an existing one, simply include them in the list returned

That sentence doesn't say anything about whether you should or shouldn't include the unchanging set of validators as well. Honestly, I'm _not sure_. To find out the answer, I guess you could look at the code, ask in a Tendermint help channel, or just do lots of tests.

kansi: @ttmc I looked in the tendermint code base [here](https://github.com/tendermint/tendermint/blob/master/state/execution.go#L251). I seems that the validator set `currentSet` is modified depending on the updates received. It seems that we don't need to return all the validators, rathers just the ones which we wish to add, modify or delete.

kansi:
> How? Just have the node operator use the MongoDB API (as local MongoDB admin user) to update the new collection (the local list of validators). No new HTTP PUT endpoint would be needed. The GET endpoint might still be useful.
The `PUT` HTTP api has been proposed so that the bdb users can write application logic to add new validators and then call the api.

> There could be some convenience CLI commands like bigchaindb add-validator PUBLIC_KEY POWER and bigchaindb change-validator-power PUBLIC_KEY POWER (for use by the node operator).

This would be a great way to quickly add validators.

kansi:
> I don't understand why we'd let external users add and remove cluster nodes (or change their power) using the HTTP API. That seems like asking for trouble.

You are probably right here. 

> Maybe "insecure version" meant something else (when we were planning this feature)?

The proposed HTTP API and implemenation details are up for discussion. There hasn't been a formal discussion about how to actually implement this feature, so we can use this rfc PR to do the same.

ttmc:
Maybe this proposal (using HTTP PUT) could be made secure by requiring that the transaction in the HTTP body be cryptographically signed:

"If the transaction is signed by validator X, then make the requested changes in the localmongodb and local Tendermint of validator X (and no other validator). Once enough of the validators get the same change, i.e. more than 2/3 of the them, then it will come into effect because that's how Tendermint works, as I understand it."

Unfortunately, I don't see a way to do an insecure version of that without asking for trouble. Maybe we could just implement the secure version and be done with it?

One concern I have is that we don't want the private key of validator X leaking out, but asking for it to be reused outside of Tendermint Core is inviting leakage. Maybe there could be a _second_ public/private keypair that's associated with the BigchainDB Server instance X (i.e. in the same BigchainDB node), and it's _that_ private key which can be used to sign the validator change messages.

kansi:
I had a discussion with @ttmc regarding the insecure and secure implementation,

It seems that the secure implementation needs due deliberation before we can implement it. Since in the secure implementation the validators are tracked in the blockcain it is natural that a validator update is proposed by an existing validator. This implies that the transaction has to signed with the validator node's key i.e. the validator node's key should be with a client which can use it to create and sign a transaction. This is potential security risk as stated by Troy

> One concern I have is that we don't want the private key of validator X leaking out, but asking for it to be reused outside of Tendermint Core is inviting leakage. 

It seems that an alternate approach is required to sign the transaction without leaking out the key to a client. @ttmc suggested maybe we could design a solution around PKI.


For the insecure version, we decided it would be enough to simply have a new sub-command `change-validators` to allows the node admins to add new validators. This should be sufficient for the MVP version.

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,103 @@
...
+  - response: 
+  ```json
+  [{"pub_key":{"type":"ed25519","data":"4E2685D9016126864733225BE00F005515200727FBAB1312FC78C8B76831255A"},
+    "power":10},
+   {"pub_key":{"type":"ed25519","data":"608D839D7100466D6BA6BE79C320F8B81DE93CFAA58CF9768CF921C6371F2553"},
+   "power":0}
+   ]
+  ```
+
+**NOTE**: when `power: 0` is set for a given validator it is removed from the validators list.
```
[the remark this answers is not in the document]
ttmc: :+1:

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Workflow overview
+Below is a summary of workflow sequence which should be executed in order to add/remove a validator dynamically,
+- BigchainDB client wishes to add/remove validators from their own node.
```
ttmc: With BigchainDB, we tend to think of "clients" as being external users interacting with the BigchainDB network via the HTTP API only. We call the person operating a node the "node operator." (I know that other blockchain software uses the word "client" to mean a node in the network.) This line could be changed to:

`- BigchainDB node operator wishes to add/update/delete a validator from their own node.`

ttmc:
Other than that minor comment about the "node operator" terminology, I think this proposal is in an implementable state.

dynamic_validators/insecure_dynamic_validators.md
```diff
@@ -0,0 +1,128 @@
...
+ "power":0}
+ ]
+```
+Note: `"power": 0` implies that the validator should be removed from the validator list. In case this diff is empty the validators remain same. Furthermore, the validators returned during `end_block` are used to update the validators of the given Tendermint node and **not** for the whole network.
+
+
+## Proposed Change
+- Add a new subcommand `bighciandb upsert-validator PUBLIC_KEY POWER` which will allow the node admin to add/update/delete a validator.
+
+- `GET /api/v1/validators` should be introduced to list the current validators.
```
ttmc: I was re-reading this spec just now and realized I don't know if this `GET /api/v1/validators` request is supposed to respond with this node's _local_ list of validators, or some kind of network-wide agreed-upon list of validators (i.e. a list that >2/3 of the nodes agree on). I suspect it's the local list. Is that right?

kansi: Yes, its the local list. I think by network-wide agreed-upon validators you mean the set of validators accepted by majority of other validators. In my view this information should be gathered by the client. This is similar to the scenario wherein a client needs to verify that a transaction was commited so it consults f+1 nodes.

ttmc:
Now that I've seen an implementation of this spec, I've realized that even though it's fairly simple, it could be even simpler. In the spirit of [keeping things simple](https://en.wikipedia.org/wiki/KISS_principle), here's a proposal for how to simplify this spec further:

- Tendermint produces new blocks quite fast, so `end_block` requests (from Tendermint core) come fairly often.
- Realistically, the node operator probably doesn't need to make several validator additions/changes/deletions per block.
- We could make our lives a lot easier by only letting the node operator made one validator addition/change/deletion per block. If they try to make more than one per block, then give them an error message and don't do it (but keep the original request pending).
- We could just use the node's local memory (RAM) to keep track of the request for a validator upsert (still made using `bigchaindb upsert-validator`). No more need for a whole new MongoDB collection, yay!
- The node operator should get some affirmative message on `stdout` or `stderr` to let them know that the `bigchaindb upsert-validator` command succeeded or failed.
- If the node crashes or the RAM gets wiped before the `end_block` request, then oh well, I guess the node operator can just try again later. No big deal. They'll know because they either got told that it failed (on `stdout` or `stderr`) or they never got any response at all. And they can use `GET /api/v1/validators` to check the list of validators.

kansi:
> We could make our lives a lot easier by only letting the node operator made one validator addition/change/deletion per block. If they try to make more than one per block, then give them an error message and don't do it (but keep the original request pending).

agreed.

> We could just use the node's local memory (RAM) to keep track of the request for a validator upsert (still made using bigchaindb upsert-validator). No more need for a whole new MongoDB collection, yay!

I would like to understand some implementation details regarding this.

> The node operator should get some affirmative message on stdout or stderr to let them know that the bigchaindb upsert-validator command succeeded or failed.

The best and definitive way for the node operator to figure our if the validator was successfully added is to the call `GET /api/v1/validators`.

ttmc:
@kansi Let's make one small change and then merge this. We can make more changes after that.

Change the file name from `dynamic_validators/insecure_dynamic_validators.md` to `3/README.md` so that it becomes basically the third BEP after 1/C4 and 2/COSS. We could refer to it informally as 3/UPSERT-VALIDATOR or similar.

I'd like to change the outline to go with what 2/COSS recommends, but that's not super high priority. This spec predates 2/COSS being finalized so that's why it's got its own outline. I guess it's still "draft"? It _has_ been implemented though...

3/README.md
```diff
@@ -1,4 +1,11 @@
-# Dynamically add/remove validators at runtime (Insecure version)
+```
+shortname: 3/UPSERT-VALIDATORS
+name: Dynamically add/remove validators at runtime (Insecure version)
+type: Feature
```
ttmc: I think the type is `Standard track`

kansi: @ttmc the same was [fixed](https://github.com/bigchaindb/BEPs/blob/master/3/README.md) in seperate PR and merged.

ttmc: Yes I noticed after making that comment. Sorry!

PR #6, opened 2018-02-15
Problem: there is no template for a spec

vrde:
Solution: "borrow" ideas from the EIP/BIP/PEP and integrate some other ideas from @muawiakh [specification document](https://github.com/bigchaindb/bigchaindb/blob/0216ffd1e8e173e1e25c2954e207d6976a60dc6/proposals/examples/example-feature-x-spec.md).

### DIFF
--- a/2/README.md
+++ b/2/README.md
@@ -50,12 +50,35 @@ A specification is a set of Markdown documents (the main file SHOULD be called `
 Every specification (including branches) carries a different number. New versions of the same specification have new numbers.
 
 ### Specification template
-TBD: Embed Muawia work here https://github.com/bigchaindb/bigchaindb/tree/master/proposals
+Each Specification MUST customize and include this header:
+````
+```
+shortname: [number/shortname]
+name: [Full name of the specification]
+type: [standard track | informational | meta ]
+status: [raw | draft | stable | deprecated | retired | deleted]
+editor: [Editor Name <email address>]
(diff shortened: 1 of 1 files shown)


2/README.md
```diff
@@ -50,12 +50,35 @@ A specification is a set of Markdown documents (the main file SHOULD be called `
...
+````
+_Note: the `number` is assigned after a specification has been submitted._
+
+Each Specification SHOULD include the following sections:
+
+1. **Abstract**. The abstract is a short (~200 word) description of the technical issue being addressed.
+
+1. **Motivation**. The motivation is critical for specifications that want to change the BigchainDB protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the specification solves. Specification submissions without sufficient motivation may be rejected outright.
+
+1. **Specification**: The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations. It MAY describe the impact on data models, API endpoints, security, performance, end users, deployment, documentation, and testing.
```
codegeschrei: Why is the impact optional?

vrde: An RFC is not necessarily technical, it can be **Informational** or **Meta**, see https://github.com/vrde/bigchaindb-rfc/blob/d87ae4a695afe4300a377a33a54f02e5cddd31e8/2/README.md#types-of-specification

PR #8, opened 2018-02-26
 Problem: RFC is a too generic name

### LINKED ISSUE #4: Rename this repository?
This repository is currently named "rfc" and the opening line on GitHub says "BigchainDB RFCs."

While using "RFC" is not _wrong_, it implies a lot of formality and expectations going back to 1969. We want something more lightweight here, I think, so we can move fast.

### Ideas

Rename this repository to:

- BigchainDB Enhancement Proposals (BEPs), similar to Python
- BigchainDB Improvement Proposals (BDBIPs, since BIPs is used already by Bitcoin)

If some of them become very stable and widely-used, then maybe they could be used as the starting point for _actual_ IETF RFCs, something like what Interledger did with some of its specs.

vrde:
Solution: switch to BigchainDB Enhancement Proposal, or BEP. Update the index page as well.

Close #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,12 @@
 
 # BigchainDB Enhancement Proposals (BEPs)
 
-This is the BigchainDB Enhancement Proposal project. We collect specifications for APIs, protocols, and processes.
+This is the BigchainDB Enhancement Proposal project. We collect BEPs for APIs, protocols, and processes.
 
-The process to add or change an EIP is the following:
-- A specification is created and modified by pull requests according to [C4](./1).
-- Specification lifecycle SHOULD follow the lifecycle defined in [COSS](./2).
-- Non-cosmetic changes are allowed only on [Raw](./2#raw-specifications) and [Draft](./2#draft-specifications) specifications.
+The process to add or change a BEP is the following:
+- A BEP is created and modified by pull requests according to [C4](./1).
+- BEP lifecycle SHOULD follow the lifecycle defined in [COSS](./2).
+- Non-cosmetic changes are allowed only on [Raw](./2#raw-beps) and [Draft](./2#draft-beps) specifications.
 
 # Current BEPs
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,5 +1,17 @@
...
+The process to add or change an EIP is the following:
+- A specification is created and modified by pull requests according to [C4](./1).
+- Specification lifecycle SHOULD follow the lifecycle defined in [COSS](./2).
+- Non-cosmetic changes are allowed only on [Raw](./2#raw-specifications) and [Draft](./2#draft-specifications) specifications.
+
+# Current BEPs
+
+Short Name | Title                                                        | Type     | Status     | Editor
+-----------|--------------------------------------------------------------|----------|------------|-------
+1/C4       | Collective Code Construction Contract                        | Meta     | Raw        | Alberto Granzotto
```
codegeschrei: Can we add links? At some point we might have more and then you need to scroll down to see a specific thing you're looking for and then scroll way (yes, at this point in time we have a lot) up to get there.

vrde: Definitely yes, good point.

README.md
```diff
@@ -1,5 +1,17 @@
 **DISCLAIMER: THIS IS A WORK IN PROGRESS**
 
-# BigchainDB RFC
+# BigchainDB Enhancement Proposals (BEPs)
 
-This is the BigchainDB RFC project.
+This is the BigchainDB Enhancement Proposal project. We collect specifications for APIs, protocols, and processes.
+
+The process to add or change an EIP is the following:
```
ttmc: EIP?

README.md
```diff
@@ -1,5 +1,17 @@
...
+- A specification is created and modified by pull requests according to [C4](./1).
```
ttmc: I wonder why you used the word "specification" here (and on the next lines) but elsewhere you were thorough in replacing all "specifications" with "BEPs".

README.md
vrde: Freudian slip :)

README.md
vrde: Thanks for pointing out, I forgot to change this file.

PR #13, opened 2018-03-11
[5/IDRP] Problem: Node operators need to know what to do about illegal data.

ttmc:
Solution: Write a BEP to reccommend something.

PR #14, opened 2018-03-12
Spec to standardize docker-compose workflows

muawiakh:
- Update base README and add this to current BEPs


ttmc:
I'm happy you are writing this up. Once we have settled on _some_ way to use Docker Compose for dev/test, we can add it to the new "Contributing to BigchainDB" docs, started with pull request  https://github.com/bigchaindb/bigchaindb/pull/2119

Note: I wanted to use BEP-4 for the Illegal Data Response Plan. Maybe you could change this one to be BEP-5. It's great that we're writing so many BEPs. Figuring out how to name them without conflict will be tricky for a while.

4/README.md
```diff
@@ -0,0 +1,207 @@
...
+- Clean up the deprecated workflow(s).
+  - Remove `docker-compose.rdb.yml`.
+  - Remove `docker-compose.benchmark.yml`.
+  - Remove `docker-compose.docs.yml`, since building documentation should be part of a development environment.
+  - Remove `docker-compose.ssl.yml`, because we are planning to remove SSL support for MongoDB.
+  - Remove `docker-compose.network.yml` file, this workflow will be supported by another deployment method. Currently, we are not using this method for development. We will have Vagrant, Ansible and K8s network driver to offer this functionality.
+
+- Update the workflows with the new deployment workflow(s).
+  - Update `docker-compose.yml` to support the current deployment strategy i.e. BigchainDB + Tendermint + Stand alone MongoDB.
+- Integrate the development and CI compose file(s) in such a way that there are minimal differences and they can be tested on a CI, also rename the files:
```
shahbazn: Why can't we use the same file for CI and dev?

4/README.md
```diff
@@ -0,0 +1,207 @@
...
+  - `docker-compose.yml` should have the base services i.e.
+  BigchainDB, MongoDB, Tendermint and any tools on top can be made part of `docker-compose.dev.yml`
+    - docker-compose.travis.yml -> docker-compose.yml
+    - docker-compose.tendermint.yml -> docker-compose.dev.yml
+
+- Update the MongoDB and Tendermint version(s) to the latest supported version:
+
+  - mongo:3.4.13
+  - tendermint/tendermint:0.13
```
shahbazn: since we are still in development phase might as well go with latest tendermint version

4/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+- [IF NEEDED]: Introduce `bigchaindb/tools` directory for tools required for development i.e.
+  - `bigchaindb-driver`
+    - Python driver to interact and transact with the BigchainDB server.
+  - `curl-client`
+    - To verify if tendermint service is up i.e. making a curl call to `/abci_query` endpoint and BigchainDB root endpoint.
+  - `nginx`
+    - For hosting docs
+
+### Technical challenges
```
ttmc: You don't need to include unused headings.

4/README.md
```diff
@@ -0,0 +1,207 @@
...
+N/A
+
+
+## Implementation
+
+### Assignee(s)
+Primary assignee(s): @muawiakh
+
+### Targeted Release
+BigchainDB Tendermint MVP
```
ttmc: Let's just call this **BigchainDB 2.0** from now on.

4/README.md
```diff
@@ -0,0 +1,207 @@
...
+Another deployment impact this change will have is with CI, we will need to update the CI scripts @`bigchaindb/.ci` to use `docker-compose.yml` instead of `docker-compose.travis.yml`.
+
+### Documentation impact
+Following documents will be impacted:
+
+- [CONTRIBUTING.md](https://github.com/bigchaindb/bigchaindb/blob/tendermint/CONTRIBUTING.md)
+- [Run BigchainDB with Docker](https://docs.bigchaindb.com/projects/server/en/tendermint/appendices/run-with-docker.html)
+- [tests/README.d](https://github.com/bigchaindb/bigchaindb/blob/tendermint/tests/README.md)
+- [TENDERMINT_INTEGRATION.rst](https://github.com/bigchaindb/bigchaindb/blob/tendermint/TENDERMINT_INTEGRATION.rst)
+
```
ttmc: + The new set of docs added to `docs/contributing/` by pull request https://github.com/bigchaindb/bigchaindb/pull/2119

4/README.md
muawiakh: I mentioned this later, because the dev will have `bigchaindb-driver`, `nginx`, `curl-client` which is used for dev only and will be an overkill for CI.

4/README.md
muawiakh: +1

4/README.md
muawiakh: Should I remove all the `N/A` headings?

muawiakh:
@ttmc Should I just mark it `5` or tackle this when any of these two is merged and the latter one can adjust?

4/README.md
shahbazn: In that case we can still keep them in the same file without CI needing to use them. e.g. `bdb` can have `mdb` and `tendermint` as dependendant services and `bigchaindb-driver` `nginx` `curl-client` can be in the same file but need separate `docker-compose up <service_name>` commands and CI can just not use them. I think it will be simpler without worrying about keeping both `docker-compose` files in sync.

4/README.md
muawiakh: I think, it is better to keep `N/A` so if someone wants to use it as reference they can check out all the things required.

muawiakh: Let me know if you think otherwise.

ttmc: Yes remove them. There's no need to keep unused headings.

4/README.md
ttmc: As I write this, I think the latest Tendermint release is actually 0.16.0 so using `latest` could get confusing, since we don't support _that_. We support the latest that `py-abci` supports, as I understand it.

4/README.md
ttmc: Yes you could add it, since it will some day point to docs about using Docker Compose to run a local node for dev/test.

ttmc:
> @ttmc Should I just mark it 5 or tackle this when any of these two is merged and the latter one can adjust?

Perhaps we could sort it out by having a _duel_. Like gentlemen. With swords of course.

4/README.md
codegeschrei: We had problems with the `latest` tendermint. The changes were not reflected in the `py-abci` and I don't know the current status of this.

muawiakh: In that case, I will stick to `0.13.0` because we are currently testing with that.

5/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+We also have multiple Dockerfiles in the repository:
+
+- Dockerfile
+  - Dockerfile used to build and publish BigchainDB production images.
+- Dockerfile-dev
+  - Dockerfile used for development purposes.
+- compose/bigchaindb-driver/Dockerfile
+  - Dockerfile used to spawn a container in development environment with bigchaindb-driver installed.
+    - This seems to be an invalid usecase for a separate container. In my opinion, `bigchaindb-driver` does not qualify as a service. It is just a python-package that can reside on a any of the containers, preferably BigchainDB.
```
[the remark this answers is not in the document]
muawiakh: On second thought, I will just remove this point. +1

muawiakh:
@vrde We don't use docker-compose for production, this will always used to development/test/quickstart scenario. We will only have one file `docker-compose.yml` which will be used for development, testing and trying out.

5/README.md
```diff
@@ -0,0 +1,166 @@
...
+- docker-compose.benchmark.yml
+  - DEPRECATED: Used to deploy standalone BigchainDB v1.3 development environment.
+
+
+Majority of the above mentioned files are either not in use OR can be integrated in a way that all of them are being tested effectively with CI pipeline(s).
+
+We also have multiple Dockerfiles in the repository:
+
+- Dockerfile
+  - Dockerfile used to build and publish BigchainDB production images.
```
shahbazn: any particular reason to keep `Dockerfile` and `Dockerfile-dev` separate?

muawiakh: Yes, because we install BigchainDB using `pip install --no-cache-dir .` in `Dockerfile` and with `Dockerfile-dev` we will use `pip install --no-cache-dir -e .[dev]` so that changes are reflected at run time.

README.md
```diff
@@ -16,3 +16,4 @@ Short Name    | Title                                                        | T
 [1/C4](1)     | Collective Code Construction Contract                        | Meta     | Raw        | Alberto Granzotto
 [2/COSS](2)   | Consensus-Oriented Specification System                      | Meta     | Raw        | Alberto Granzotto
 [3/UPSERT-VALIDATORS](3) | Dynamically add/remove validators at runtime (Insecure version)  | Standard | Raw | Vanshdeep Singh
+[5/STANDARIDIZE-DC](5) | Standard process to set up a local node for development & testing, using Docker Compose | Standard | Raw | Muawia Khan
```
vrde: The number should be `4`: `shortname: 4/STANDARIDIZE-DC`

muawiakh: I kept it 5 because @ttmc had a comment that his PR https://github.com/bigchaindb/BEPs/pull/13, is numbered 4 and we might have a conflict. I can change, whatever you guys suggest. @ttmc @vrde

5/README.md
```diff
@@ -0,0 +1,165 @@
+```
+shortname: 4/STANDARIDIZE-DC
```
vrde: Typo: `STANDARDIZE`

muawiakh:
part of bigchaindb/bigchaindb#2125

Issue #10, opened 2018-03-05
Problem: Unclear what a "Public Contract" is in this context


ttmc:
In 1/C4, [the section on "Evolution of Public Contracts"](https://github.com/bigchaindb/BEPs/tree/master/1#26-evolution-of-public-contracts) isn't clear what it means by "Public Contracts".

It _seems_ to be talking about BigchainDB Enhancement Proposals, BEPs, but it says that nowhere.

Or is a "Public Contract" something else? Is so, what?

If it's a BEP, can we just change the name in 1/C4 from "Public Contracts" to "BEPs" for clarity and consistency? We could even link to 2/COSS.

vrde:
You are right, it's not clear.

From my understanding, a "Public Contract" should be equivalent to our [Public API](https://github.com/bigchaindb/bigchaindb/blob/master/CHANGELOG.md#change-log-release-notes) plus all our BEPs.

Does it make sense to you as well?

ttmc:
Oh, our Public API... we can include that in the Public Contracts, yes.

Would it make sense to add a paragraph to the section on "Evolution of Public Contracts" to list the things included in the "Public Contracts": All BEPs, our Public API, maybe other things?

We could make a small BEP that specifies "what is in the BigchainDB Public API." Do you think that would make sense? It's weird having the Public API defined in the header of CHANGELOG.md.

ttmc:
Some day, the HTTP API, the Events API, the BigchainDB Server CLI, the transaction model(s), and more can all be in BEPs.

We already have the transaction model as a nice spec document. I just need to find time to move that over to become BEPs (one BEP per transaction model version, I think). They are currently over at https://github.com/ipdb/ipdb-tx-spec

vrde:
> Some day, the HTTP API, the Events API, the BigchainDB Server CLI, the transaction model(s), and more can all be in BEPs.

Agree, but this might take ∞ amount of time :fire_engine: 

What about improving section _2.1. Preliminaries_ with those new items:
- A "Public Contract" is a contract between the project and the end-user. It can be a BEP, an API, a Command Line Interface, or anything the end-user rely on.
- The project SHALL use [Semantic Versioning](https://semver.org/) (note: Public API is a Public Contract).

ttmc:
To resolve this issue, see the comment:

https://github.com/bigchaindb/BEPs/pull/15#issuecomment-372589957

PR #15, opened 2018-03-12
Problem: The BigchainDB Public API isn't defined

ttmc:
Solution: Write a BEP to define the BigchainDB Public API.

As a follow-up once this PR is merged, `CHANGELOG.md` over in `bigchaindb/bigchainb` could be updated to remove this definition. It didn't really belong there anyway.


7/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Documentation of the Current BigchainDB Public API
+
+At the time of writing, the various components of the BigchainDB Public API were documented in various places. Those were:
+
+* [Configuration settings (latest release)](https://docs.bigchaindb.com/projects/server/en/latest/server-reference/configuration.html)
+* [Command Line Interface (latest release)](https://docs.bigchaindb.com/projects/server/en/latest/server-reference/bigchaindb-cli.html)
+* [Transaction Spec (all versions)](https://github.com/ipdb/ipdb-tx-spec)
+* [HTTP API (latest release)](https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html)
+* [Events API (latest release)](https://docs.bigchaindb.com/projects/server/en/latest/events/index.html)
```
vrde: Do you think we should link to a specific version of the documentation at one point? (where *at one point* is when we have the documentation for BigchainDB 2.0 online).

vrde:
Will this PR resolve #10?

7/README.md
ttmc: That is a good idea, but right now there are no URLs for BigchainDB 2.0, e.g. on ReadTheDocs, because it doesn't exist yet. It will come from a Git Tag on the commit which _defines_ BigchainDB 2.0, which will also be tied to the GitHub _release_. That Git Tag doesn't exist yet, but it will soon! And then we can make links to all the BigchainDB 2.0 specific specs. Yay!

ttmc:
> Will this PR resolve #10?

Not _completely_. We will still have to say what we _mean_ by "public contracts" in the 1/C4 document itself, to give the reader complete clarity. Quoting from _Social Architecture_ page 93, which explains that line:

> By "public contracts", I mean APIs and protocols.

That's quite clear! I would like to add some text about where _BigchainDB's_ APIs and protocols are specifically defined, i.e. a link to BEP-7 as a start. Which needs to exist first, so I can link to it.

vrde:
> That's quite clear! I would like to add some text about where BigchainDB's APIs and protocols are specifically defined, i.e. a link to BEP-7 as a start. Which needs to exist first, so I can link to it.

Perfect, can you take care of it?

PR #20, opened 2018-03-14
Problem: No BEP discussing the UTXO implementation

kansi:
Solution: Write an informational BEP with a brief discussion of current UTXO implementation.

### DIFF
--- a/utxo-info/README.md
+++ b/utxo-info/README.md
@@ -1,0 +1,19 @@
+```
+shortname: ?/UTXO-IMPL
+name: A short discription of the current UTXO implementation
+type: informational
+status: Draft
+editor: Vanshdeep Singh <vanshdeep@bigchaindb.com>
+```
+
+## Description
+The UTXO (unspent transaction output) of each transaction is tracked using the a merkle tree. The merkle root of this tree is used as the app hash and retuned to Tendermint during `commit`. Below is a short summary of the UTXO is calculated,
+
+- Each unspent output is considered as a leaf node in the merkle tree.
+- The nodes are added/removed with each incoming transaction i.e. during `commit` when the transactions are being processed for bulk write the UTXO are updated in the `utxo` collection.
+- Once all the transactions in a block are processed and stored, all the objects in the `utxo` collection are fetched and stored i.e. a list of leaves is prepared.
+- The merkle root is calculated using this list of leaves.
+
+
+## Copyright Waiver
+To the extent possible under law, the person who associated CC0 with this work has waived all copyright and related or neighboring rights to this work.


utxo-info/README.md
```diff
@@ -0,0 +1,19 @@
+```
+shortname: ?/UTXO-IMPL
+name: A short discription of the current UTXO implementation
```
ttmc: description

utxo-info/README.md
```diff
@@ -0,0 +1,19 @@
...
+type: informational
+status: Draft
+editor: Vanshdeep Singh <vanshdeep@bigchaindb.com>
+```
+
+## Description
+The UTXO (unspent transaction output) of each transaction is tracked using the a merkle tree. The merkle root of this tree is used as the app hash and retuned to Tendermint during `commit`. Below is a short summary of the UTXO is calculated,
```
ttmc: It's not clear how all the outputs (which are like floating independent dots) are somehow connected to form a Merkle tree. Maybe draw a picture? They say pictures are worth 1000 words and in this case I think it's true.

ttmc: Maybe link to some articles about Merkle trees.

https://en.wikipedia.org/wiki/Merkle_tree

What about AVL trees, IAVL trees, and IAVL+ trees. I gather that those are supposed to be useful somehow, as an improvement. Is it the rules for adding the next leaf UTXO, where to put it?

kansi: @ttmc I have restrained the context of this BEP to just discussing the current implementation. Also, the leaves of the UTXO merkle tree are added to the list of the existing leaves and the list is sorted after which merkle root is calculated. As explained in the BEP, the leaves are added to the `utxo` collection and then all the leaves are read back and sorted, after which the root is calculated.

utxo-info/README.md
```diff
@@ -0,0 +1,25 @@
...
+The UTXO (unspent transaction output) of each transaction is tracked using the a [merkle tree](https://en.wikipedia.org/wiki/Merkle_tree). The merkle root of this tree is used as the app hash and returned to Tendermint during `commit`. Below is a short summary of how the UTXO is calculated,
+
+- Each unspent output is considered as a leaf node in the merkle tree.
+- The unspent outputs are stored in `utxo` collection in MongoDB.
+- The unspent outputs are added/removed from the `utxo` collection with each incoming transaction i.e. during `commit` when the transactions are being processed for bulk write the UTXO are updated in the `utxo` collection.
+- Once the transactions in a block are processed and stored, all the objects in the `utxo` collection are fetched.
+- This list of unspent outputs fetched from the `utxo` collection is considered as the leaves of the UTXO merkle tree. Refer the image below,
+- The list of leaves is hashed and sorted after which the merkle root is calculated by pairing adjacent leaves.
+
+![Merkel Tree](https://upload.wikimedia.org/wikipedia/commons/thumb/9/95/Hash_Tree.svg/800px-Hash_Tree.svg.png)
```
ttmc: Merkle

utxo-info/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+
```
ttmc: It's still not clear to me _where_ one puts a new leaf node (UTXO). There are many possible spots to choose from. And it matters. The choice will affect the final hash. It seems that's where the AVL rules come into use. Is this right?

kansi: Okay, lets take a step back. The unspent outputs or UTXO is stored in the `utxo` collection in MongoDB (there is no insert order here). When the merkle root is calculated all the leaves are fetched from the `utxo` collection, the leaves are hashed (lets call this `H`), then this hash list is sorted (`Hs`) and the merkle root is caculated by pairing the leaves. So if a new leaf is inserted we don't particularly insert it to specific position in the tree, its position among leaves will be determinted by the `sort` function which sorts the list of hashes `H`.

ttmc: Hmm, seems like we are relying on some mysterious `sort` function. It would be nice to know what it does and how it works. Can we reliably get a nice balanced tree doing this?

utxo-info/README.md
vrde: :de: :evergreen_tree:

utxo-info/README.md
kansi: @ttmc You can have look at the code [here](https://github.com/bigchaindb/bigchaindb/blob/tendermint/bigchaindb/tendermint/lib.py#L162). The list of `hashes` are `sorted` and the merkle root is calculated. Since the whole tree is built everytime the merkle root is calculated so it would be balanced.

ttmc: Oh wow, rebuilding the whole tree from scratch every time _will_ make it balanced, but that's a real waste of computing time and power. It's okay for now, but we need to do it smart in the future, using a proper IAVL+ library. That library should have a method like `tree.insert_node(new_node)` and then there will be other methods to compute the Merkle root of the tree (i.e. the app hash. That will also be fast, because most of those hash computations were done before and are stored somewhere handy).

kansi: @ttmc You are right. We need to have some discussions around how to improve this implementation and make it crash recovery friendly.

vrde:
Should we move it to an issue and label it with **BEP idea**?

PR #21, opened 2018-03-14
Problem: Unclear who does what

ttmc:
Solution: List the C4 roles and who does them, if known


ttmc:
The list of Maintaners seems too short, but I don't want to get in trouble. Add yourself if you are a maintainer of something. For example, we need some maintainers (official) for the Python and JavaScript drivers.

PR #27, opened 2018-03-20
Problem: BigchainDB Transaction Spec docs should live under bigchaindb/BEPs

ttmc:
Solution: Move them there!

I also added some boilerplate / template outline stuff from COSS.

I also merged the `changes.md` file right into the spec document for BigchainDB Transaction Spec v2.0, to make it so there's only one text file.

All instances of the word "IPDB" got edited to something else, usually BigchainDB.


ttmc:
Thanks @vrde

PR #28, opened 2018-04-05
Problem: BEP-3 not really all that "insecure"

ttmc:
Solution: Update descriptions of BEP-3 so it's more consistent with the actual security of BEP-3.
          
- Also update the name of the target release to BigchainDB 2.0.
- Also update the status of BEP-3 from "raw" to "stable".

### DIFF
--- a/3/README.md
+++ b/3/README.md
@@ -1,16 +1,15 @@
 ```
 shortname: 3/UPSERT-VALIDATORS
-name: Dynamically add/remove validators at runtime (Insecure version)
+name: Dynamically add/update/remove validators at runtime
 type: standard
 status: raw
 editor: Vanshdeep Singh <vanshdeep@bigchaindb.com>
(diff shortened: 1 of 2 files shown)


3/README.md
```diff
@@ -1,16 +1,15 @@
 ```
 shortname: 3/UPSERT-VALIDATORS
-name: Dynamically add/remove validators at runtime (Insecure version)
+name: Dynamically add/update/remove validators at runtime
 type: standard
 status: raw
```
vrde: Status is `raw` here, but `stable` in the main `README.md` table. My suggestion is to keep this proposal `draft` for now, and mark it `stable` when BigchainDB 2.0 is not alpha anymore, what do you think?

ttmc: Oops, good catch! I'll fix that.

PR #29, opened 2018-04-05
Problem: intro is not the first thing to read

vrde:
Solution: rearrange sections and put intro first


README.md
```diff
@@ -1,5 +1,29 @@
 **DISCLAIMER: THIS IS A WORK IN PROGRESS**
 
+# BigchainDB Enhancement Proposals (BEPs)
```
ttmc: Do we need that top line disclaimer about this being a work in progress?

vrde: IMO no :smile:

PR #31, opened 2018-04-10
Problem: BigchainDB 2.0 has design issues

vrde:
Solution: Share ideas on how to move forward


codegeschrei:
Hey, great work!
For the image we use in there. Can we rename `Drv` to `Driver`? I think there is enough space to just write it out. And people don't have to guess if it is `Driver` or maybe something else.
Just had to look up all the [possible meanings](https://www.abbreviations.com/DRV) of `Drv` :D

vrde:
> For the image we use in there. Can we rename Drv to Driver? I think there is enough space to just write it out. And people don't have to guess if it is Driver or maybe something else.

Done, thanks :)

10/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+BigchainDB 2.0 Alpha 1 is now the composition of three different system (MongoDB, Tendermint, and BigchainDB itself). This setup generates different problems: a) managing multiple services together adds overhead and complicates deployment; b) the current design makes some calls going from BigchainDB to Tendermint to BigchainDB again, penalizing performance; c) there is data duplication between LevelDB and MongoDB; d) we need to maintain [`py-abci`][pyabci], while the official implementation of the [ABCI][abci] interface is in Golang; e) the logic to validate transactions is spread across multiple files and it's difficult to manage, and a heavy refactoring is needed; f) the configuration is spread across BigchainDB and Tendermint.
+
+In this BEP, we discuss a possible path towards an architecture more suitable for BigchainDB 2.0.
```
ttmc: This paragraph, and those that follow, aren't really about the "Motivation" for the BEP any more. Maybe you could insert a new section header here, such as "Overview".

10/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+In this BEP, we evaluate a rewrite of BigchainDB in another language, specifically Golang. The benefits are multiple: Golang it is a fairly simple language, easy to learn and read; it is shown to be efficient and faster than Python and can compile to a binary file; it is used in several big projects, and it's the language of choice for major blockchain projects: this simplifies code reuse if we need it in the future, and allows us a tighter integration with Tendermint.
+
+Rewriting a system to another language is a risky and possibly endless process. To minimize the risk, an incremental approach is proposed. The approach is incremental in the following sense:
+- It is divided into different subsequent steps.
+- Every step is self contained and outputs a usable system.
+- Every step should be a unit of work reasonably small.
```
ttmc: This would read better as:
```text
- Every step should be a reasonably small unit of work.
```

10/README.md
```diff
@@ -0,0 +1,88 @@
...
+Posting a transaction to BigchainDB is one of the main primitives of the system.
+
+The goal of _Step 1_ is to expose through `bigchaindb-go` one single endpoint fully compatible with [`POST /api/v1/transactions`][bdb:post-tx], and implement the logic required to satisfy the [ABCI][abci] interface.
+
+The integration between the existing BigchainDB Server code (written in Python) and the new `bigchaindb-go` is described in the following section.
+
+About the existing BigchainDB Server:
+- All the HTTP API endpoints will still be served from the existing code.
+- The endpoint `POST /api/v1/transactions` will be just a _pass through proxy_ to the `bigchaindb-go` implementation.
+- A new endpoint `POST /api/v1/validate` will be implemented, to check the validity of a transaction (this **temporary** endpoint will be used by `bigchaindb-go` to do the actual validation, and will give us more time to actually focus on the experiment).
```
ttmc: Shouldn't this be a `GET /api/v1/validate` endpoint, that is, using an HTTP GET method, since it's just to get information, not to post or store some new information.

Also, I wonder why it should be a temporary endpoint. It might be useful for outside users of the BigchainDB network, to find out of a transaction would be valid if it were POSTed at the instant the request was made.

10/README.md
```diff
@@ -0,0 +1,88 @@
...
+- It implements the ABCI interface (`check_tx`, `begin_block`, `deliver_tx`, `end_block`, `commit`), and communicates to the local MongoDB instance to write transactions once they are validated, ideally using the [in process interface][abci:in-process].
+
+The following diagram shows how the two systems work together.
+
+![Interaction between the existing BigchainDB Server and bigchaindb-go][diagram]
+
+It is important to notice that the existing BigchainDB Server will only **read** from the shared MongoDB instance, while `bigchaindb-go` will be the only process able to **write** data in the underlying database.
+
+### Other steps
+If the experiment is successful, other API endpoints can be easily integrated one by one into the `bigchaindb-go` implementation. This BEP does not cover this part for now.
```
ttmc: I wonder if we ever need to translate the other endpoints into Golang. They're mostly for doing special queries via GET requests. I've been writing about how node operators can expose their local MongoDB database to the outside world, and their options for doing so. They _could_ just to expose MongoDB on port 27017 (to authenticated users with limited access permissions, such as read-only). But a node operator could also expose an HTTP API or a GraphQL API to their MongoDB database, for example. The "HTTP API for Querying BigchainDB" could be a _stand alone_ program / executable / package. It would be up to the node operator to decide what to expose to external users, and how.

10/README.md
ttmc: Regarding question 1), I realized why you chose to use the HTTP POST method: it's because GET requests aren't supposed to have a HTTP request body, and some web servers put a limit on how big the URI can be, so if you used GET, there would be nowhere to stuff the transaction! Maybe explain that in the document.

10/README.md
```diff
@@ -0,0 +1,88 @@
+```
+shortname: 10/CBD
+Name: Consolidating BigchainDB 2.0 Design
```
ttmc: This title, `Consolidating BigchainDB 2.0 Design` seems to broad and generic. If I read that title, and nothing else, in a year from now, I won't have any idea what the BEP is about. What are the stand-out features of this BEP?

- Golang
- It uses the "Strangler Application" approach to rewriting the code
- It's more about rewriting the code than changing the design. In some ways the high-level design stays the same, because as a black box, it should _do the same thing_ before and after the rewrite.

An alternate title might be, `A Strangler Application Approach to Rewriting Some Code in Go`

ttmc: The short name would also have to be updated.

10/README.md
vrde: > Shouldn't this be a GET /api/v1/validate endpoint, that is, using an HTTP GET method, since it's just to get information, not to post or store some new information.

and

> Regarding question 1), I realized why you chose to use the HTTP POST method: it's because GET requests aren't supposed to have a HTTP request body, and some web servers put a limit on how big the URI can be, so if you used GET, there would be nowhere to stuff the transaction! Maybe explain that in the document.

That's a valid point, but it's *controversial* to have a GET with a body, see:
- https://stackoverflow.com/a/983458/597097
- https://www.elastic.co/guide/en/elasticsearch/guide/current/_empty_search.html

> Also, I wonder why it should be a temporary endpoint. It might be useful for outside users of the BigchainDB network, to find out of a transaction would be valid if it were POSTed at the instant the request was made.

This is true, I'll remove "temporary". If anyone wants to write a BEP around it and make it part of the public interface, they are welcome :slightly_smiling_face:.

10/README.md
vrde: Yep, what you are saying makes total sense. This BEP just mention that integration is easy, but does not actively promote it.

10/README.md
```diff
@@ -1,20 +1,21 @@
...
-# Consolidating BigchainDB 2.0 Design
+# A Strangler Application Approach to Rewriting Some Code in Go
 
 ## Abstract
 This document describes an experimental architecture for BigchainDB Server, illustrated in the diagram below. It replaces a small amount of Python code with Go code, allowing the BigchainDB team to experiment with Golang with low risk while achieving several other goals, including easier migration to new versions of Tendermint and higher performance.
 
 ## Motivation
 BigchainDB 2.0 Alpha 1 has been developed using BigchainDB 1.3 as a starting point. Design decisions that were correct for the first generation of BigchainDB, are now challenged by the integration of a new consensus algorithm (managed by Tendermint). Specifically, the Tendermint state machine runs in a separate process, and communicates with BigchainDB using multiple channels.
 
+## Overview
```
ttmc: I would move this subsection heading (**Overview**) to below the following paragraph, since the following paragraph lists a bunch of motivating factors. That is, move it just before the "In this BEP, we..." paragraph.

ttmc:
@vrde would it be okay if I merged this PR, or are you waiting for something else first?

vrde:
> @vrde would it be okay if I merged this PR, or are you waiting for something else first?

I'd merge it :dancer:

shahbazn:
This is already merged but I still want to add my opinion here. 

- I do like the idea of slowly moving to go implementation as it will massively improve the packaging and distribution of our software and it will definitely simplify the configuration e.g. It will be really awesome if 3.0 release is a single binary that our users can just download and easily run with the deployment system of their choice. 
- Since we don't have any benchmarks for current performance stats the term `higher performance` seems a little abstract. We also don't know what the bottlenecks are for us be it communication between tendermint and bigchaindb services or the performance of python vs go in specific use cases.
- I am not sure if shifting towards go implementation will remove the LevelDB and MongoDB data duplication. 
- I do like the idea of tight integration with the census system of our choice i.e tendermint and removing the overhead of maintaining py-abci
- It will help us have do a better job at code organization now that we have better process and contribution guidelines.
- It raises the question of putting efforts on refactoring current python code vs slowly moving the code to go

I think this should be considered for implementation after we are done with 2.0 release as a broader shift towards go based implementation with clear motivations or else we might end up increasing overhead of maintaining another library without any quantifiable benefits.

ttmc:
@shahbazn - You wrote:

> I am not sure if shifting towards go implementation will remove the LevelDB and MongoDB data duplication.

Indeed, that's probably a task for a separate BEP, but this BEP will change how that BEP gets implemented, so they're related.

> It raises the question of putting efforts on refactoring current python code vs slowly moving the code to go

This BEP keeps all transaction-construction and -validation code in Python, for now, and that's where the refactoring would happen. Even if we eventually translate _that_ code to Go, it's probably a good idea to refactor it in Python first, since that's the language we all know and we can do it quickly / soon. See my proposed [BEP-9](https://github.com/bigchaindb/BEPs/pull/33) for a list of simple things we could do to get started refactoring that transaction-related code.

vrde:
> I do like the idea of slowly moving to go implementation as it will massively improve the packaging and distribution of our software and it will definitely simplify the configuration e.g. It will be really awesome if 3.0 release is a single binary that our users can just download and easily run with the deployment system of their choice.

:+1: 

> Since we don't have any benchmarks for current performance stats the term higher performance seems a little abstract. We also don't know what the bottlenecks are for us be it communication between tendermint and bigchaindb services or the performance of python vs go in specific use cases.

True, it was an assumption based on the fact that a tighter integration would lead to less roundtrips, and to the fact that Golang is generally faster than Python.

> I am not sure if shifting towards go implementation will remove the LevelDB and MongoDB data duplication.

Let's say that moving to Golang opens up this possibility.

PR #35, opened 2018-04-30
Problem: drivers reliability is low

vrde:
Solution: write a proposal on how drivers make use of all nodes in the network.


vrde:
> LGTM. I think Sylvain already allowed for a list of nodes in the Python driver but never implemented that case. Whoever implements this BEP can see what's already in place.

Thanks! Yes, I remember he put some logic there, and hopefully we can use part of it.

PR #37, opened 2018-05-02
Problem: the BEP template does not ask to specify solution criteria

ldmberman:
Solution: include the "Solution Criteria" section into the BEP template


ttmc:
I think that talking about "solution criteria" is a bit abstract and hard to imagine. Maybe you could instead say something like, "a detailed explanation of the problem, along with details of what it would take to solve it."

vrde:
That short memo is really interesting! Changing this BEP might take a bit of time, since it's part of the "costitution". I have some ideas to share with you, I need few more days.

ttmc:
@ldmberman Yes I like the idea of a **Problem Breakdown** section. Would it replace the proposed **Solution Criteria** section?

ldmberman:
@ttmc yes, it would be a replacement.

ttmc:
@ldmberman Okay, maybe make that change and see what @vrde says.

PR #43, opened 2018-05-16
Problem: BEP-10 goals are vague.

ldmberman:
Solution: Be very explicit about BEP-10 goals. Additionaly, be more explicit about choosing Go as a language. Add some minor refinements.

### DIFF
--- a/10/README.md
+++ b/10/README.md
@@ -4,21 +4,60 @@ Name: A Strangler Application Approach to Rewriting Some Code in Go
 type: Informational
 status: Raw
 Editors: Alberto Granzotto <alberto@bigchaindb.com>
-Contributors: Vanshdeep Singh <vanshdeep@bigchaindb.com>
+Contributors: Vanshdeep Singh <vanshdeep@bigchaindb.com>, Lev Berman <ldmberman@gmail.com>
 ```
 
 # A Strangler Application Approach to Rewriting Some Code in Go
 
 ## Abstract
-This document describes an experimental architecture for BigchainDB Server, illustrated in the diagram below. It replaces a small amount of Python code with Go code, allowing the BigchainDB team to experiment with Golang with low risk while achieving several other goals, including easier migration to new versions of Tendermint and higher performance.
+This document describes an experimental architecture for BigchainDB Server, illustrated in the diagram below. It replaces a small amount of Python code with Go code, allowing the BigchainDB team to experiment with Golang with low risk and at the same time make the step forward the long-term goals of the increased transaction rate and better platform reliability.
 
 ## Motivation
 BigchainDB 2.0 Alpha 1 has been developed using BigchainDB 1.3 as a starting point. Design decisions that were correct for the first generation of BigchainDB, are now challenged by the integration of a new consensus algorithm (managed by Tendermint). Specifically, the Tendermint state machine runs in a separate process, and communicates with BigchainDB using multiple channels.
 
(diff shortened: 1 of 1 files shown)


10/README.md
```diff
@@ -4,29 +4,70 @@ Name: A Strangler Application Approach to Rewriting Some Code in Go
...
 
-BigchainDB 2.0 Alpha 1 is now the composition of three different system (MongoDB, Tendermint, and BigchainDB itself). This setup generates different problems: a) managing multiple services together adds overhead and complicates deployment; b) the current design makes some calls going from BigchainDB to Tendermint to BigchainDB again, penalizing performance; c) there is data duplication between LevelDB and MongoDB; d) we need to maintain [`py-abci`][pyabci], while the official implementation of the [ABCI][abci] interface is in Golang; e) the logic to validate transactions is spread across multiple files and it's difficult to manage, and a heavy refactoring is needed; f) the configuration is spread across BigchainDB and Tendermint.
+BigchainDB 2.0 Alpha 1 is now the composition of three different system (MongoDB, Tendermint, and BigchainDB itself).
+
+This setup generates different problems:
+
+1. managing 3 services together adds operational overhead and complicates distribution;
+
+    Blockchain clients are commonly independent binaries. This is true for Bitcoin, Etherium, and Monero.
```
ttmc: Ethereum

10/README.md
```diff
@@ -4,29 +4,70 @@ Name: A Strangler Application Approach to Rewriting Some Code in Go
...
+
+2. the current design raises certain performance concerns;
+
+    The TCP communications between BigchainDB and Tendermint, the transaction validation implemented in Python, and Crypto-Conditions implemented in Python might significantly contribute to the performance degradation. The detailed benchmarks are on the way.
+
+3. there is data duplication between LevelDB and MongoDB;
+
+    By tightly integrating with Tendermint, the default database backend can be replaced.
```
ttmc: I'm not sure tight integration is needed before the LevelDB database can be replaced by the MongoDB database. All one would have to do is make a custom version of Tendermint Core, i.e. by changing the code of Tendermint Core, which is written in Go. Learning Go (and the internals of Tendermint Core) seems like the main thing blocking us from replacing LevelDB with MongoDB. The BEP helps us start learning Go.

10/README.md
ttmc: Ethereum

10/README.md
ldmberman: > All one would have to do is make a custom version of Tendermint Core, i.e. by changing the code of Tendermint Core, which is written in Go

You actually described what I am referring to by "tight integration". The term is probably not very intuitive.. I will try to change this part.

ldmberman:
In general, if we agree on what we have and do not plan to add much to this BEP, I can also make it "Draft" here.

ttmc:
Please don't make it "Draft" until we've done some of the things specified in the BEP. To see how BEP-2 (COSS) defines "draft", see:

https://github.com/bigchaindb/BEPs/tree/master/2#draft-beps

10/README.md
ttmc: I don't understand but it doesn't matter right now. We can think about the whole "replace LevelDB" idea/design in another BEP.

ldmberman:
I understand how it works for "standard" proposal, not that much for "informational" ones.

10/README.md
```diff
@@ -4,29 +4,70 @@ Name: A Strangler Application Approach to Rewriting Some Code in Go
...
 ## Overview
-In this BEP, we evaluate a rewrite of some parts of BigchainDB in another language, specifically Golang. The benefits are multiple: Golang it is a fairly simple language, easy to learn and read; it is shown to be efficient and faster than Python and can compile to a binary file; it is used in several big projects, and it's the language of choice for major blockchain projects: this simplifies code reuse if we need it in the future, and allows us a tighter integration with Tendermint.
+In this BEP, we evaluate a rewrite of some parts of BigchainDB in Golang.
+
+The main reason to choose Golang is:
+
+* Tendermint is written in Golang. Using Go allows to achieve the most efficient integration with Tendermint. It includes efficient ABCI communications, interactions with the database, and straightforward compilation of BigchainDB and Tendermint into a single binary.
+
+
+Other notes about Go:
```
kansi: I would like to understand the reason behind writing these notes. IMHO the document should only have reason for leaning towards Golang in comparison to python. So, I think these notes should be contrasted with python.

ldmberman:
@kansi I refined the description of the motivation behind choosing the language.

PR #44, opened 2018-05-22
[WIP] Problem: Validators cannot vote on proposals (BEP-18)

vrde:
Solution: write a transactional election protocol.

### DIFF
--- a/18/README.md
+++ b/18/README.md
@@ -14,12 +14,11 @@ contributors: Vanshdeep Singh <vanshdeep@bigchaindb.com>, Lev Berman <lev@bigcha
 This specification introduces the new concept of **Election**. An Election is an asynchronous process that when successful triggers Network wide changes synchronously (i.e. at the same *block height*). An Election is started by any Validator in the Network, called **Initiatior**. The Election itself and all cast Votes are transactions, hence stored in the blockchain. This enables new Validators to replay Network changes incrementally, while syncing.
 
 ## Motivation
-Changing the shape of a BigchainDB Network at runtime is an important requirement. While this BEP does not address this issue directly, it wants to solve the limitations we had with [Upsert Validators (BEP-3)][BEP-3] by giving a tool that can be used this and other situations. BEP-3 addresses a really interesting use case: adding a Member to a running Network. Since changes in the Validator Set are not themselves expressed as transactions but are included into the metadata of an arbitrary block, all the nodes have to add a validator simultaneously—if less than the majority of the nodes attempt to include a Validator, the block fails and the whole Network can stuck. Since those changes are **not** stored in the blockchain itself, a new Node that needs to sync with the Network will fail in validating blocks from a specific *height* on. The *height* will match the one of the first block voted by a new Validator that was **not** included in the `genesis.json` file (and Tendermint will fail with `Invalid commit (last validator set size) vs (number of precommits)`).
+Changing the shape of a BigchainDB Network at runtime is an important requirement. While this BEP does not address this issue directly, it wants to solve the limitations we had with [Upsert Validators (BEP-3)][BEP-3] by giving a tool that can be used this and other situations. BEP-3 addresses a really interesting use case: adding a Member to a running Network. Since changes in the Validator Set are not themselves expressed as transactions but are included into the metadata of an arbitrary block, all the nodes have to add a Validator simultaneously—if less than the majority of the nodes attempt to include a Validator, the block fails and the whole Network can stuck. Since those changes are **not** stored in the blockchain itself, a new Node that needs to sync with the Network will fail in validating blocks from a specific *height* on. The *height* will match the one of the first block voted by a new Validator that was **not** included in the `genesis.json` file (and Tendermint will fail with `Invalid commit (last Validator set size) vs (number of precommits)`).
 
 The abstract version of the problem is "make Network wide decisions". We need a way for Members to vote on proposals, and implement them as soon as the quorum of ⅔ is reached. In this BEP, a proposal is named *Election*, a vote is a fungible token formalized with the capitalized name *Vote Token*. Elections and Vote Tokens are stored in the blockchain, so a new Node that syncs with the Network will be able to replay them locally, and reach the same state of all other Members.
 
-The basic idea is to formalize the concept of an Election storing its data in a [BigchainDB Transaction Spec v2 (BEP-13)][BEP-13].
-By storing it in a BigchainDB Network, it allows Members to vote asynchronously, and Validators to apply changes synchronously.
+The basic idea is to formalize the concept of an Election storing its data in a [BigchainDB Transaction Spec v2 (BEP-13)][BEP-13]. Implementations of this BEP might introduce new operations for a *BigchainDB Transaction Spec v2*, and new validation rules for the Election transaction and the Vote transaction. By storing it in a BigchainDB Network, it allows Members to vote asynchronously, and Validators to apply changes synchronously.
 
 ## Specification
 At any point in time, a Member of a BigchainDB Network can start a new Election.This Member is called **Initiator**.
@@ -32,10 +31,10 @@ During the `end_block` call, all transactions about to be committed are checked.
 
 Each Validator checks every new transaction that is about to be committed in a block. The process is roughly the following:
 1. If the transaction is **not** a valid Vote, return.
-2. If `asset.data.id` is **not** a valid Election, return.
-3. If the Election has less than ⅔ of positive votes, return.
-4. If the Election has been already implemented, return.
-5. Since all previous checks has been successful, execute the logic to implement the Election.
+2. If the Vote is for a **not** valid Election, return.
+3. If the Election **including** the current Vote has less than ⅔ of positive votes, return.
+4. If the Election **excluding** the current Vote has more than ⅔ of positive votes, return. (It has been concluded already.)
+5. If the Election **excluding** the current Vote has less than ⅔ of positive votes and **including** the current Vote has more than ⅔ of positive votes, execute the logic to implement the Election.
 
 ### What is a valid Vote?
 A Validator must be able to discern valid Votes from invalid ones. A valid vote is a transaction that spends one or more Vote Tokens to the Election Address.
@@ -50,19 +49,19 @@ A Validator must be able to discern valid Elections from invalid ones. A valid E
 5. Each Validator is represented in `outputs`.
 6. Each entry in `outputs` can be spent by only one Validator, and the amount attached to it is equal to the power of that Validator.
 
-**Note: any change in the Validator Set will make old Elections invalid.**
+**Note: any change in the Validator Set will make old Elections invalid. Check [approach 2](#generalized-approach-approach-2) for a process that can tolerate a certain degree of change to the Validator Set.**
 
 ### Extra: Vote delegation
 Vote delegation is trivial. Let's consider a Network of three Members: Alice, Bob, and Carly. Alice is the Initiator, and starts a new Election. Alice generates a `CREATE` transaction with three outputs, one per each Member. Bob wants to delegate his vote to Carly, so he transfers his output to Carly, granting her more votes she can spend in the way she wants.
 
-### Constraints
+### About Election finality
 Let the Validator Set be denoted by ![V_t][eq_V_t] at time ![][eq_t], a member of a BigchainDB Network can start a new Election. This member is called **Initiator** (![v_it][eq_v_it] s.t. ![v_it in V][eq1]) .
 
 #### Valid Election
-An Election proposal is a transaction representing the matter of change. The _Initiator_ ![v_it][eq_v_it], issues a create election transaction ![][EC_t]. The election MUST contain outputs  such that each output is populated with the public key of the validator ![][eq_v_k] s.t. ![][v_k_in_V_t], and an amount equal to the current power ![][eq_p(v_k, t)]  of the validator.
+An Election proposal is a transaction representing the matter of change. The _Initiator_ ![v_it][eq_v_it], issues a create election transaction ![][EC_t]. The election MUST contain outputs  such that each output is populated with the public key of the Validator ![][eq_v_k] s.t. ![][v_k_in_V_t], and an amount equal to the current power ![][eq_p(v_k, t)]  of the Validator.
 
 #### Voting
-Once ![][EC_t] is committed in a block, the election starts. Independently and asynchronously, each validator may spend its vote tokens (referred as ![][T_k]) to the election address ![][EC_t_addr] to show agreement on the matter of change. The Election Address ![][EC_t_addr] is the `id` of the transaction ![][EC_t].
+Once ![][EC_t] is committed in a block, the election starts. Independently and asynchronously, each Validator may spend its vote tokens (referred as ![][T_k]) to the election address ![][EC_t_addr] to show agreement on the matter of change. The Election Address ![][EC_t_addr] is the `id` of the transaction ![][EC_t].
 
 NOTE: Once the vote tokens ![][T_k] have been transferred to the election address ![][EC_t_addr] it is not possible to transfer it again, because the private key is not known.
 
@@ -74,7 +73,7 @@ At time ![][t_n] let,
 
 
 ##### Constrained approach (Approach 1)
-In the constrained approach any change to the validator set invalidates all the elections which were initiated with a different validator set.
+In the constrained approach any change to the Validator Set invalidates all the elections which were initiated with a different Validator Set.
 
 If below conditions hold true then the election is concluded and the proposed change is applied,
 1. ![][V_tn_equals_V_t]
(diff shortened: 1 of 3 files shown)


18/README.md
```diff
@@ -0,0 +1,208 @@
...
+      ]
+    }
+  ]
+  "version": "2.0"
+}
+```
+
+## Rationale
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+Another idea to implement this protocol is to use a multi signature transaction.
```
ttmc: The Rationale section should explain why the above design was chosen rather than some other design. In this case, you should explain why a multi-signature transaction wasn't used.

18/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Backwards Compatibility
+<!--All BEPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The BEP must explain how the author proposes to deal with these incompatibilities. BEP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+This BEP is fully backwards compatible.
+
+## Implementation
+<!--The implementations must be completed before any BEP is given status "stable", but it need not be completed before the BEP is accepted. While there is merit to the approach of reaching consensus on the BEP and rationale before writing code, the principle of "rough consensus and running code" is still useful when it comes to resolving many discussions of API details.-->
+
```
ttmc: You can say that this section can be filled in once there _is_ an implementation.

18/README.md
```diff
@@ -0,0 +1,208 @@
...
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations. It MAY describe the impact on data models, API endpoints, security, performance, end users, deployment, documentation, and testing.-->
+
+At any point in time, a Member of a BigchainDB Network can start a new Election.This Member is called **Initiator**.
+
+An Election is a transaction representing the matter of change, and some Vote tokens. The Initiator issues a `CREATE` transaction with the `amount` set to the total number of Validators, and transfer one vote token per Validator using the transaction `outputs`.
+
+At this point the Election starts. Independently, and asynchronously, each Validator can spend its Vote Token to an Election Address to show agreement on the matter of change. The Election Address is the `id` of the first `CREATE` transaction. Once a Vote Token has been transferred to that address, it is not possible to transfer it again, because there private key is not known.
```
ttmc: I don't understand how you can transfer/spend an election token to the 'id' of a CREATE transaction. That will be a hash, but in BigchainDB, one transfers/spends outputs to public keys, not hashes.

kansi:
After discussing this spec with @vrde following issue should be addressed i.e.

The `CREATE` transaction proposed in the current spec allocates one vote to each node in the network. This implies that all the nodes in the network have an equal influence on the election. Another factor which should be taken in account is power of the nodes i.e. each node in the network has a power (`integer`) associated to it which is taken into account when calculating if the majority (`>2/3`) concensus has been achieved. This power should be reflected in the `CREATE` transaction i.e. each validator should be allocated a number of votes equal to their power in the network. So, the `CREATE` transaction is only included/commited in a block if each of nodes in the network aggree to the number of votes allocated to the each of nodes for the election. Furthermore, when votes are being casted using `TRANSFER` transaction each node should cross verify the number of votes a given node is allowed to cast by comparing the current validator and the `CREATE` transaction. Any mismatch should result in the vote being discarded.

18/README.md
```diff
@@ -0,0 +1,208 @@
...
+          "public_key": "Election Address",
+          "type": "ed25519-sha-256"
+        },
+        "uri": "ni:///sha-256;b...123?fpt=ed25519-sha-256&cost=131072"
+      },
+      "public_keys": [
+        "Election Address"
+      ]
+    }
+  ]
```
kansi: Comma is missing

18/README.md
```diff
@@ -0,0 +1,208 @@
...
+2. If `metadata.type` is **not** `vote`, return false.
+3. If `inputs.owners_before` is **not** a Validator, return false.
+5. Return true.
+
+### Validate Election
+A Validator must be able to discern valid Elections from invalid ones. The process is roughly the following:
+1. `type`: must contain the string `election`.
+2. `name`: name of the election.
+3. `version`: version number.
+4. `matter`: a human readable, short paragraph on the matter of change.
```
kansi: The field is not specified in the schema below.

18/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+# Transactional Election Process
+
+## Abstract
+<!-- The abstract is a short (~200 word) description of the technical issue being addressed. -->
+This specification introduces the new concept of **Election**. An Election is an asynchronous process that when successful triggers Network wide changes synchronously (i.e. at the same *block height*). An Election is started by any Validator in the Network, called **Initiatior**. The Election itself and all cast Votes are transactions, hence stored in the blockchain. This enables new Validators to replay Network changes incrementally, while syncing.
+
+## Motivation
+<!--The motivation is critical for BEPs that want to change the BigchainDB protocol. It should clearly explain why the existing protocol BEP is inadequate to address the problem that the BEP solves. BEP submissions without sufficient motivation may be rejected outright.-->
+Changing the shape of a BigchainDB Network at runtime is an important requirement. While this BEP does not address this issue directly, it wants to solve the limitations we had with [Upsert Validators][BEP-3] by giving a tool that can be used this and other situations.
```
ldmberman: > it wants to solve the limitations we had with [Upsert Validators][BEP-3]

I would add a more detailed explanation of the limitations.

As I see them:

- Since Validator Set changes are not themselves expressed as transactions but are included into the metadata of an arbitrary block, all the nodes have to add a validator simultaneously - if less than the majority of the nodes attempt to include a validator, the block fails. Not only the attempt to add a validator fails but all the transactions in the block have to go through the consensus process again. The proposed algorithm, on the other hand, consists of the individual transactions each node can verify without making itself a decision to include a validator.
- Nodes fail to replay the blocks committed by the validators they do not know yet. The `Invalid commit (last validator set size) vs (number of precommits)` problem.

18/README.md
vrde: Yes, I'll add it soon.

18/README.md
vrde: That's a tricky part I should explain more. Last night I wrote a [note about that](https://github.com/vrde/notes/tree/master/bigchaindb#using-a-public-key-as-a-topic).

A public key can be generated from a private key, or initialized with 32 bytes (256 bits). The `id` of a transaction is a 256 bit long hash. So we can use `id`s as seeds to initialize a public key :champagne:. I need to expand this section to explain this process. Maybe we can have a quick chat with @kansi and you. It can be an `informal` BEP, so it can be reused in other `BEP`s.

18/README.md
```diff
@@ -0,0 +1,212 @@
...
+Changing the shape of a BigchainDB Network at runtime is an important requirement. While this BEP does not address this issue directly, it wants to solve the limitations we had with [Upsert Validators (BEP-3)][BEP-3] by giving a tool that can be used this and other situations. BEP-3 addresses a really interesting use case: adding a Member to a running Network. Since changes in the Validator Set are not themselves expressed as transactions but are included into the metadata of an arbitrary block, all the nodes have to add a validator simultaneously—if less than the majority of the nodes attempt to include a Validator, the block fails and the whole Network can stuck. Since those changes are **not** stored in the blockchain itself, a new Node that needs to sync with the Network will fail in validating blocks from a specific *height* on. The *height* will match the one of the first block voted by a new Validator that was **not** included in the `genesis.json` file (and Tendermint will fail with `Invalid commit (last validator set size) vs (number of precommits)`).
```
kansi: .. whole Network can `get` stuck ..

18/README.md
```diff
@@ -15,7 +15,9 @@ This specification introduces the new concept of **Election**. An Election is an
...
```
ldmberman: > all the nodes have to add a validator simultaneously—if less than the majority of the nodes attempt to include a Validator, the block fails and the whole Network can stuck. Since those changes are **not** stored in the blockchain itself

I would put it slightly differently:

1. all the nodes have to agree on the validator set updates;
2. if the majority agrees on the update, the majority can further post transactions but all other nodes who did not agree on the update get stuck.

If the network is small, there might be not enough nodes to form a majority and thus the whole network might be stuck but it's fine - it just reflects the disagreements between the nodes on the validator topic.

> Since those changes are **not** stored in the blockchain itself, a new Node that needs to sync with the Network will fail in validating blocks from a specific *height* on.

They are actually stored in the blockchain. If the node does not accept the validator, this behaviour is expected. If the node does accept the validator, it needs to determine the height when to include it upon replay. Basically, that's the main problem the election approach solves - determines the height when to include the new validator.

18/README.md
```diff
@@ -0,0 +1,212 @@
...
+status: raw
+editor: Alberto Granzotto <alberto@bigchaindb.com>
+```
+
+
+# Transactional Election Process
+
+## Abstract
+<!-- The abstract is a short (~200 word) description of the technical issue being addressed. -->
+This specification introduces the new concept of **Election**. An Election is an asynchronous process that when successful triggers Network wide changes synchronously (i.e. at the same *block height*). An Election is started by any Validator in the Network, called **Initiatior**. The Election itself and all cast Votes are transactions, hence stored in the blockchain. This enables new Validators to replay Network changes incrementally, while syncing.
```
muawiakh: nit: maybe a `,` after `successful, triggers network wide changes`

18/README.md
muawiakh: Maybe add a reference to `BEP/19` here, I feel I can, understand the motivation more clearly when I understand whats happening in `BEP/19`  and over here it is kind of skimmed .

ldmberman:
I have some concerns about implementing proposals and votes via assets compared to new transaction types.

As of now:

- users have almost full control over the asset structure - it's completely up to them what they store inside the `"data"` field;
- BigchainDB only makes sense of the transaction version, there is no such thing as an asset version.

Here I am trying to ponder about what we will face if we change either of those.

Do we validate proposal transactions upon their placement? If yes, we need to put additional effort into implementing asset validation and we further restrict what users can put into their assets. If not, we lose the convenient feature of reporting about the invalid proposals early on.

In any case, since assets are parts of transactions, introducing the election mechanism requires a new version of the transaction spec - it needs to be documented there that certain assets gain special treatment.

Versioning itself poses another question - if we support multiversioned transactions, when do we increase the transaction version and when do we increment the election asset version? The whole topic is quite vague at the moment, but IMO those versions might be connected and we may end up writing validators that check for a cartesian product of transaction and asset versions to apply different validation.

I understand the desire to build on top of what we already have but on the other hand network changes is a fundamental entity so it is worth making them first-class citizens of the system.

I think, it's possible to implement and manage it both ways, but due to the abovementioned I am personally in favor of the new transaction types, like `NETWORK_UPDATE` and `NETWORK_UPDATE_VOTE`.

What do you think guys?

ldmberman:
@ttmc a new transaction type is a backwards-compatible spec change so we can do `2.1`. Also, we need to add a new transaction spec version even if we go on with special assets.

kansi:
I concur i.e. it would be good to solve this problem via new transaction types i.e. `NETWORK_UPDATE` and `NETWORK_UPDATE_VOTE`. So a summary of what changes should be done  is as follows,
- Extend `CREATE` transaction sepc and define `NETWORK_UPDATE` transaction sepc and write the corresponding BEP. Some tasks here would be to write spec for `asset.data`, restrict crypto-conditions to `ed25519` etc.
- Extend `TRANSFER` transaction spec and define `NETWORK_UPDATE_VOTE` transaction sepc and write the corresponding BEP.
- Once we settle on the above-described specs, then refactor/add validation logic for new proposed sepcs in the server code.
- Once the server has been updated the drivers should be updated to allow the creation of these new transaction types (js, python).

This approach is much elaborate than the current one and its been already suggested that beta releases should be feature freezed so I don't think this approach can be implemented before stable release. So I see two options here,

- We rollback the `upsert-validator` feature altogether as the current implementation doesn't work that smoothly and work on the above approach after the stable release.

- Have the current implementation via `asset.data` and later upgrade to above-suggested approach.

@vrde @ttmc @ldmberman  let me know which approach would be preferable.

ldmberman:
Insteand of rolling `upsert-validator` back, we can consider including the complete current set of validators into every `EndBlock` to make the current implementation usable.

The implementation with `asset.data` also requires an update of the transaction spec.

ttmc:
> The implementation with `asset.data` also requires an update of the transaction spec.

The "[BigchainDB Transactions Spec v2](https://github.com/bigchaindb/BEPs/tree/master/13)" allows `asset.data` to be almost any JSON (with a few limitations imposed by MongoDB). If we give special meanings to some values of `asset.data`, then I'm not sure that belongs in the "transactions spec"; it seems out of scope. It can go in other specs, such as BEP-18.

ldmberman:
>  If we give special meanings to some values of asset.data, then I'm not sure that belongs in the "transactions spec"; it seems out of scope.

@ttmc how do you decide what to include into the transaction spec? Why should the spec describe the special meaning of `version`, `outputs`, `inputs`, and `asset` keys but not describe the special meaning of the `asset.data.type` and `asset.data.version` keys?

18/README.md
```diff
@@ -2,211 +2,158 @@
...
-3. `version`: version number.
-4. `matter`: a human readable, short paragraph on the matter of change.
-5. If `inputs.owners_before` is **not** a Validator, return.
-6. If every `outputs.owners_after.amount` is **not** exactly `1`, return.
-7. If `outputs.owners_after` does **not** contain exactly all Validators, return.
-8. Apply any other validation rule enforced by the code specific to the election itself, return false if it fails.
-9. Return true.
+2. If `asset.data.id` is **not** a valid Election, return.
+3. If the Election has less than ⅔ of positive votes, return.
+4. If the Election has been already implemented, return.
```
ldmberman: What do you think about replacing `has been already implemented` with `already has more than ⅔ of positive votes`, for clarity?

It would then expand into 3 bullet points instead of 2:
```
- If the Election including the current vote has less than ⅔ of positive votes, return.
- If the Election excluding the current vote has more than ⅔ of positive votes, return. It has been concluded already.
- If the Election excluding the current vote has less than ⅔ of positive votes and including the current vote has more than ⅔ of positive votes, conclude the election.
```

18/README.md
```diff
@@ -2,211 +2,158 @@
...
+The abstract version of the problem is "make Network wide decisions". We need a way for Members to vote on proposals, and implement them as soon as the quorum of ⅔ is reached. In this BEP, a proposal is named *Election*, a vote is a fungible token formalized with the capitalized name *Vote Token*. Elections and Vote Tokens are stored in the blockchain, so a new Node that syncs with the Network will be able to replay them locally, and reach the same state of all other Members.
 
-The basic idea is to formalize the concept of an Election storing its data in a [BigchainDB Transaction Spec v2][BEP-13].
+The basic idea is to formalize the concept of an Election storing its data in a [BigchainDB Transaction Spec v2 (BEP-13)][BEP-13].
```
ldmberman: Should we mention the need for a new version of the transaction spec here?

18/README.md
```diff
@@ -2,211 +2,158 @@
...
-      },
-      "public_keys": [
-        "Daniel's public key"
-      ]
-    }
-  ],
-  "version": "2.0"
-}
-```
+### Constraints
```
ldmberman: I think, it makes sense to put this into the `About Election finality` section - the text there seems to be a perfect introduction into what we explain here.

18/README.md
```diff
@@ -2,211 +2,158 @@
...
+
+##### Approach 1 vs Approach 2
+
+Approach 1 is easier to comprehend and explain because of how constrained it is.
+
+Approach 2 has two benefits over the Approach 1.
+
+1. Elections can have intersections. In other words, the following is possible:
+
+```election 1 creation height < election 2 creation height < election 1 conclusion height < election 2 conclusion height```
```
ldmberman: I was not right about the intersections - in case 2 elections do not update the validator set, they can interleave even in the first approach. So, the second approach is just about surviving validator set updates.

18/README.md
```diff
@@ -2,211 +2,158 @@
...
+A Validator must be able to discern valid Elections from invalid ones. A valid Election is a transaction where **all** the following conditions are true.
+
+1. `operation` is `CREATE`, or equivalent.
+2. Each public key listed in `inputs.owners_before` belongs to one of the Validators.
+3. `outputs.amount` is the total power of the Validators.
+4. `outputs` has as many entries as the total number of Validators.
+5. Each Validator is represented in `outputs`.
+6. Each entry in `outputs` can be spent by only one Validator, and the amount attached to it is equal to the power of that Validator.
+
+**Note: any change in the Validator Set will make old Elections invalid.**
```
ldmberman: It seems to belong to the `What is a valid Vote?` section. The vote is valid if it transfers some vote tokens to a valid election and the validator set has not changed since the election was placed. Does it sound good?

PR #45, opened 2018-05-23
Problem: Upsert validator approach is error prone (BEP-21)

kansi:
Solution: Propose an alternate approach


3/3-1.md
```diff
@@ -0,0 +1,94 @@
...
+- At this point the node `N6` won't see `N5` as a validator because the `genesis.json` would only contain `{N1,N2,N3,N4}` as validators.
+- Since `N5` was added dynamically and `upsert-validator` only mutates the local validator set of a given node it implies that in order for the `N6` to see `N5` as a validator it would have to execute `upsert-validator N5_PUB_KEY N5_POWER` on its own node at the exact block height (when syncing with the network) when `N5` was previously added by the network.
+
+
+From the above description it is evident that propagating dynamically added validators is a major hassle with huge possibility of errors.
+
+
+## Technical details
+
+To solve the aforementioned issue, following change to the behaviour of `upsert-validator` is being proposed. We use the transaction election process (TEP) proposed in [BEP 18](https://github.com/bigchaindb/BEPs/pull/44) to form a to automate and synchronize the operation on the validator set.
```
ttmc: This sentence got mangled somehow:

"...to form a to automate and synchronize the operation on the validator set."

3/3-1.md
```diff
@@ -0,0 +1,94 @@
...
+    "version": "1.0",
+    "args": {
+        "public_key": "Wn2DedV9OA0LJJjOxr7Sl7jqCSYjQihA6dCBX+iHaEI=",
+        "power": 10,
+        "node_id": "82190eb6396bdd80b83aef0f931d0f45738ed075"
+    }
+}
+```
+
+NOTE: The `CREATE` transaction is signed using the private key generated and stored by Tendermint in `priv_validator.json`.
```
ttmc: You should clarify that this is node A's Tendermint private key (an Ed25519 private key).

3/3-1.md
```diff
@@ -0,0 +1,94 @@
...
+4. If the node operator aggrees to the operation being proposed by the `node_upsert_request_id` then they can vote on the same using the following,
+
+```
+$ bigchaindb upsert-validator approve-request node_upsert_request_id
+```
+
+The above command `POST`s a `TRANSFER` transaction casting the vote of the node for the given TEP.
+NOTE: The `TRANSFER` transaction is signed using the private key generated and stored by Tendermint in `priv_validator.json`.
+
+
```
ttmc: Maybe there should be a **Backwards Compatibility** section.

ttmc: :+1:

3/3-1.md
```diff
@@ -0,0 +1,98 @@
+```
+shortname: 3.1/UPSERT-VALIDATORS
```
ttmc: I understand why you want to number this 3.1: it replaces BEP-3, but BEPs have integer numbers.

Once this BEP is implemented, it will become stable, and then we can deprecate BEP-3 and add a note in the BEP-3 text telling readers that it was replaced by this BEP, which can have a number like BEP-20.

vrde: > I understand why you want to number this 3.1: it replaces BEP-3, but BEPs have integer numbers.
> 
> Once this BEP is implemented, it will become stable, and then we can deprecate BEP-3 and add a note in the BEP-3 text telling readers that it was replaced by this BEP, which can have a number like BEP-20.

I'd add a hyperlink from BEP-3 to BEP-20. We can formalize this process in BEP-2.

19/transaction_network_election_v2.0.yaml
```diff
@@ -0,0 +1,53 @@
...
+  # version should greater than or equal the base transaction spec version
+  # - major number should be equal to the base transaction spec major version
+  # - minor number should be greater than or equal to base transaction spec minor version
+  version:
+    type: string
+    pattern: "^2\\.0$"
+  asset:
+    additionalProperties: false
+    properties:
+      data:
```
ldmberman: `asset.data` is historically defined by users, not by the spec. What do you think about putting the spec keys into `asset`?

19/transaction_network_election_v2.0.yaml
```diff
@@ -0,0 +1,53 @@
...
+- operation
+- version
+- asset
+- outputs
+properties:
+  operation: "VALIDATOR_ELECTION"
+  # version should greater than or equal the base transaction spec version
+  # - major number should be equal to the base transaction spec major version
+  # - minor number should be greater than or equal to base transaction spec minor version
+  version:
```
ldmberman: I think, multiple versions make sense only when there are moving parts. Since "VALIDATOR_ELECTION" and "CREATE" transactions will always be processed by the same soft, it makes sense to use a single version for them.

19/transaction_network_election_v2.0.yaml
```diff
@@ -0,0 +1,53 @@
...
+"$schema": "http://json-schema.org/draft-04/schema#"
+type: object
+title: Network Election Schema - Propose a network change
+required:
+- operation
+- version
+- asset
+- outputs
+properties:
+  operation: "VALIDATOR_ELECTION"
```
ldmberman: I like the idea of specific naming since we do not have a clear picture about other election use-cases yet. Can you also update the title and the file name?

19/transaction_network_election_v2.0.yaml
kansi: I concur, as of now, the version is same as the `CREATE` transaction spec. The only reason we might need a minor version greater than the `CREATE` spec version number is when the `asset.data` spec is changed for whatever reason. I don't see that happening anywhere in the future but I have left room for such possibility.

19/transaction_network_election_v2.0.yaml
kansi: That would lead to unnecessary code when separating `asset` and `metadata` from a transaction.

ldmberman:
> I don't see that happening anywhere in the future but I have left room for such possibility.

It's not only the room for a possibility but also the burden of updating the version in 2 places whenever the `CREATE` or `TRANSFER` version is updated. Therefore if we do not see any benefits from having 2 versions, we need to keep one.

19/transaction_network_election_v2.0.yaml
ldmberman: Discussed offline: to preserve the semantics of `asset.data` across transactions of different type, we want to consider including these fields into the transaction object itself and storing them in a separate collection for efficient indexes and small transaction document size.

ldmberman:
I'll move it here - the previous comment was collapsed.

Discussed offline with @kansi: to preserve the semantics of `asset.data` across transactions of different type, we want to consider including these fields into the transaction object itself and storing them in a separate collection for efficient indexes and small transaction document size.

Also discussed the conditions to include a validator w. r. t. ever changing validator set:

- `VALIDATOR_ELECTION` transactions created with the amount of outputs less than the current validator set are rejected;
- multiple TEP can go on simultaneously; new validators can not vote on transactions created prior to them joining but as long as the TEP set constitutes a supermajority the new validator can be added;
- `VALIDATOR_ELECTION` transactions allocating more than 1/3 of the current total voting power are rejected. Tendermint won't accept such transactions anyway (currently it panics).

ldmberman:
Below is the summary of the discussion we had with @kansi about multiple elections recieving the last vote(s) in the same block.

Let's say we are building a block. Consider:

`T` - total validator power before the block is being built.
`Pi` - the new power proposed by the `i`th election.
`Vi` - the total power of validators voted on the `i`th election.

The following has to be true in order for the `i`th election to be concluded (we check it on every `DeliverTx` and `EndBlock`):
```
Pi + Pi-1 + ... + P0 < 1/3T
Vi > 2/3 (T + Pi-1 + ... + P0)
```

The first requirement is for security. If it is not satisfied, Tendermint will reject the validator set update.

The second requirement can be actually relaxed to `Vi > 2/3T`. The additional `Pi-1 + ... + P0` are there to make it consistent with the situation when all these elections are concluded in the separate blocks `Bi-1`, ..., `B0`.

There can be a situtation when the `i`th election satisfies the second requirement, but does not satisfy the first one. In this case the transaction will be rejected and the election might never be concluded unless the voter revotes. We need to further think if we keep it as-is for now or can think of a clean way to address this problem.

19/README.md
```diff
@@ -0,0 +1,91 @@
...
+shortname: [19/UPSERT-VALIDATORS]
+name: Dynamically add/update/remove validators at runtime
+type: standard
+Status: raw
+editor: Vanshdeep Singh <vanshdeep@bigchaindb.com>
+```
+
+## Description
+
+The current [BEP 3](https://github.com/bigchaindb/BEPs/blob/master/3/README.md) spec for adding new validators requires that the node administrators coordinate when dynamically adding a new validator i.e. all the node operators have to simultaneously execute `upsert-validator` in order to ensure that the validator is added on all the nodes. A lack of coordination can result in a situation wherein only a part of the network updates their validator set which could eventually lead to a network hangup. Furthermore, for a sufficiently large network the coordination itself can become a tedious task.
```
vrde: > The current [BEP 3](https://github.com/bigchaindb/BEPs/blob/master/3/README.md)

In the link, use the relative path `../3`.

19/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+From the above description it is evident that propagating dynamically added validators is a major hassle with huge possibility of errors.
+
+
+## Technical details
+
+To solve the aforementioned issue, following change to the behaviour of `upsert-validator` is being proposed. We use the transaction election process (TEP) proposed in [BEP 18](https://github.com/bigchaindb/BEPs/pull/44) to automate and synchronize the operation on the validator set.
+
+Consider a network of 4 nodes `{A,B,C,D}` if now if a node A wishes to add a new node `E` to the network then following steps should be followed,
```
vrde: > if now if a node

Check again this sentence :)

19/README.md
```diff
@@ -0,0 +1,91 @@
...
+- [BEP 18](https://github.com/bigchaindb/BEPs/pull/44)
+- [Tendermint validators](http://tendermint.readthedocs.io/en/master/specification/validators.html)
+
+
+## Copyright Waiver
+To the extent possible under law, the person who associated CC0 with this work has waived all copyright and related or neighboring rights to this work.
+
+
+[spec_validator_election]: ./transaction_network_election_v2.0.yaml
+[spec_validator_election_vote]: ./transaction_network_election_vote_v2.0.yaml
```
vrde: The two links are incorrect.

19/README.md
```diff
@@ -0,0 +1,91 @@
...
+## Technical details
+
+To solve the aforementioned issue, following change to the behaviour of `upsert-validator` is being proposed. We use the transaction election process (TEP) proposed in [BEP 18](https://github.com/bigchaindb/BEPs/pull/44) to automate and synchronize the operation on the validator set.
+
+Consider a network of 4 nodes `{A,B,C,D}` if now if a node A wishes to add a new node `E` to the network then following steps should be followed,
+
+1. Node `A` executes
+
+```
+$ bigchaindb upsert-validator E_PUBKEY E_POWER E_NODE_ID --private-key /home/user/.tendermint/config/priv_validator.json
```
vrde: Given `TOTAL_POWER` as the sum of all powers of all current validators, should the command return an error **if not** `0 < E_POWER < TOTAL_POWER / 3`?

19/README.md
```diff
@@ -0,0 +1,91 @@
...
+<election_id>
+```
+
+The above command `POST`s a [`VALIDATOR_ELECTION`][spec_validator_election] transaction and returns the `election_id`.
+
+NOTE:
+- The `VALIDATOR_ELECTION` transaction is signed using the private key generated and stored by Node `A`'s' Tendermint in `priv_validator.json`.
+
+
+2. The `election_id` is then manually sent (via email or message) to rest of the nodes in the network.
```
vrde: Instead of _to the rest of the nodes in the network_, what about _to all members of the network_? I'd like to make a distinction between humans and machines.

19/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+```
+$ bigchaindb upsert-validator E_PUBKEY E_POWER E_NODE_ID --private-key /home/user/.tendermint/config/priv_validator.json
+<election_id>
+```
+
+The above command `POST`s a [`VALIDATOR_ELECTION`][spec_validator_election] transaction and returns the `election_id`.
+
+NOTE:
+- The `VALIDATOR_ELECTION` transaction is signed using the private key generated and stored by Node `A`'s' Tendermint in `priv_validator.json`.
```
vrde: What about:
> The `VALIDATOR_ELECTION` transaction is signed with the private key of the Validator proposing the election. The private key is generated by Tendermint, and stored in the `priv_validator.json` in the Tendermint home.

19/README.md
```diff
@@ -0,0 +1,91 @@
...
+power=10
+node_id=82190eb6396bdd80b83aef0f931d0f45738ed075
+```
+
+The above command list the details about the node which is being added/updated/deleted from the network.
+
+4. If the node operator aggrees to the operation being proposed by the `election_id` then they can vote on the same using the following,
+
+```
+$ bigchaindb upsert-validator approve-request node_upsert_request_id --private-key /home/user/.tendermint/config/priv_validator.json
```
vrde: It's not clear that `node_upsert_request_id` is a parameter. Should it be called `<election_id>` instead?

vrde: I see there are three commands:
- `bigchaindb upsert-validator <validator data>`
- `bigchaindb upsert-validator show-request <election_id>`
- `bigcahindb upsert-validator approve-request <election_id>` 

To simplify the CLI, I was thinking of:
- `bigchaindb upsert-validator create <validator data>`
- `bigchaindb upsert-validator show <election_id>`
- `bigcahindb upsert-validator approve <election_id>`

19/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Assignee(s)
+Primary assignee(s): @kansi
+
+### Targeted Release
+BigchainDB 2.0
+
+
+## Reference(s)
+- [BEP 3](https://github.com/bigchaindb/BEPs/blob/master/3/README.md)
```
vrde: Please use the relative path `../3`.

19/README.md
kansi: If we use the name `upsert-validator create` it is not immediately clear that the sub-command also updates and deletes validators.

19/README.md
```diff
@@ -0,0 +1,93 @@
...
+<election_id>
+```
+
+The above command `POST`s a [`VALIDATOR_ELECTION`][spec_validator_election] transaction and returns the `election_id`.
+
+NOTE:
+- The `VALIDATOR_ELECTION` transaction is signed with the private key of the validator proposing the election. The private key is generated by Tendermint, and stored in the `priv_validator.json` in the Tendermint's `$TMHOME` directory (usually located at `~/.tendermint/`).
+
+- `E_POWER` should be such that `0 < E_POWER < TOTAL_POWER/3` where `TOTAL_POWER` denotes the total voting power of the current network.
+
```
ldmberman: Makes sense to put it here that there has to be an output for every validator in the network.

19/transaction_validator_election_vote_v2.0.yaml
```diff
@@ -0,0 +1,31 @@
+---
+"$schema": "http://json-schema.org/draft-04/schema#"
+type: object
+title: Validator Election Vote Schema - Vote on a validator set change
+required:
+- operation
+- version
+- outputs
```
ldmberman: What do outputs of the vote transaction describe?

kansi: To whom the cost is being casted to. The voter can choose to delegate their vote or vote to the `election_id`.

19/README.md
kansi: A general description of the election process was missing for I addressed that and your concern here
https://github.com/bigchaindb/BEPs/pull/45/files#diff-8e56cd5a0d784d78d68efd53fa940687R69

vrde:
@ldmberman wrote:
> Discussed offline with @kansi: to preserve the semantics of asset.data across transactions of different type, we want to consider including these fields into the transaction object itself and storing them in a separate collection for efficient indexes and small transaction document size.

In my opinion, the "election fields" should be stored in `asset.data`, and eventually enforced by the new specific schemas. If we keep using `asset.data` we inherit the logic to split the asset from the transaction itself, without the need to create a specific collection for the "election fields" **and** write custom code to handle that.

> Also discussed the conditions to include a validator w. r. t. ever changing validator set:
>  -   VALIDATOR_ELECTION transactions created with the amount of outputs less than the current validator set are rejected;

The `amount` of the transaction MUST be the amount of total power of the validators. The amount MUST be spent fully in the `outputs` of the `VALIDATOR_ELECTION` transaction. Every single output amount MUST match the power of the validator it has been transferred to.

>  -   multiple TEP can go on simultaneously; new validators can not vote on transactions created prior to them joining but as long as the TEP set constitutes a supermajority the new validator can be added;

Multiple `TEP`s can go simultaneously, but when one succeeds, all the others are invalid, because the amounts in the other `VALIDATOR_ELECTION` transactions don't match the current set of validators.

>  -   VALIDATOR_ELECTION transactions allocating more than 1/3 of the current total voting power are rejected. Tendermint won't accept such transactions anyway (currently it panics).

:+1:

ldmberman:
>  If we keep using asset.data we inherit the logic to split the asset from the transaction itself, without the need to create a specific collection for the "election fields" and write custom code to handle that.

Do you also want to store the election parameters in the same collection? It means users will see the election data inside their full text search queries.

> The amount of the transaction MUST be the amount of total power of the validators. The amount MUST be spent fully in the outputs of the VALIDATOR_ELECTION transaction. Every single output amount MUST match the power of the validator it has been transferred to.

That's a great way to put it!

> Multiple TEPs can go simultaneously, but when one succeeds, all the others are invalid, because the amounts in the other VALIDATOR_ELECTION transactions don't match the current set of validators.

There are 2 ways of handling it. I advocate for not invalidating the elections as long as the conditions I described [here](https://github.com/bigchaindb/BEPs/pull/45#issuecomment-398825857) are met. It does not sacrifice security but improves user experience significantly. Election is a time-consuming multi-party process. Repeating it is a burden worth avoiding IMO.

vrde:
>> If we keep using asset.data we inherit the logic to split the asset from the transaction itself, without the need to create a specific collection for the "election fields" and write custom code to handle that.

> Do you also want to store the election parameters in the same collection? It means users will see the election data inside their full text search queries.

Full text search queries are anyway returning *not necessarily relevant data* to the user. Anyone can *interfere* with the result of a query by posting transactions with specific keywords. Also, a voting transaction is *rare* compared to all other transactions (plus, if a user wants to search for an election, they can do it).

>> Multiple TEPs can go simultaneously, but when one succeeds, all the others are invalid, because the amounts in the other VALIDATOR_ELECTION transactions don't match the current set of validators.

> There are 2 ways of handling it. I advocate for not invalidating the elections as long as the conditions I described here are met. It does not sacrifice security but improves user experience significantly. Election is a time-consuming multi-party process. Repeating it is a burden worth avoiding IMO.

For me it is simpler (and probably more secure?) to handle sequential elections. If we keep the rules I stated before:
> The `amount` of the transaction MUST be the amount of total power of the validators. The amount MUST be spent fully in the `outputs` of the `VALIDATOR_ELECTION` transaction. Every single output amount MUST match the power of the validator it has been transferred to.

then if an election succeeds, all other open elections are now **invalid** elections, because the set of validators changed. In my opinion this approach is simpler because we only have to keep in the logic the **current** set of validators. If we have to handle multiple concurrent elections we *might* have to keep a different state for all of them in order to validate them.

ldmberman:
> Full text search queries are anyway returning not necessarily relevant data to the user.

Fair enough.

> In my opinion this approach is simpler because we only have to keep in the logic the current set of validators

Ah, I was always keeping the current set of validators in mind otherwise it's too complicated indeed. 

We have 2 types of criteria in the algotihtm - the criteria of an election transaction that is considered valid **and** the criteria of a concluded election.

> The amount of the transaction MUST be the amount of total power of the validators. 

If we make this a requirement for the valid election transaction but not the requirement for a concluded transaction, we can conclude as many of elections as possible as long as each of them meets the conclusion criteria that only depend on the the current state.

There is a little bit of extra complexity to track the state when multiple of the concluded elections are in the same block but the difference is basically `Vi > 2/3 (T + Pi-1 + ... + P0)` vs `Vi > 2/3 T` what's not a big complexity increase IMO.

vrde:
> If we make this a requirement for the valid election transaction but not the requirement for a concluded transaction, we can conclude as many of elections as possible as long as each of them meets the conclusion criteria that only depend on the the current state.

OK, now I see what you mean: if a transaction does not satisfy the condition
> The amount of the transaction MUST be the amount of total power of the validators.

then the transaction is rejected and **not** stored in the blockchain. So a `VALIDATOR_ELECTION` is stored in the blockchain if and only if it's valid. So yes, you don't need to keep a *state* because everything that is in the blockchain itself is already valid. Yep it seems easier than what I thought at the beginning.

Thinking about corner cases (capital letters represent nodes):
- Genesis with `A, B, C, D, E`
- `A` pushes a VALIDATOR-ELECTION `e₁` to remove `E` (upsert-validator with power `0`)
- `E` pushes a VALIDATOR-ELECTION `e₂` to add `F`
- `E` transfers its vote for `e₂` to `Alice`
- `A, B, C, D` vote positively for `e₁`, and E gets kick out of the network

Is in your opinion `e₂` still valid? It's an open question, I don't have a strong opinion on that at the moment.

ldmberman:
According to the criteria discussed, `e₂` is still valid. Why not?

Reclaiming all the actions made by a removed validator (like initiated elections or cast votes) is a separate feature we might consider in the future.

kansi:
> Full text search queries are anyway returning not necessarily relevant data to the user. Anyone can interfere with the result of a query by posting transactions with specific keywords. Also, a voting transaction is rare compared to all other transactions (plus, if a user wants to search for an election, they can do it).

Agreed. In view of the above the argument I don't think there is any other motivation to have system-level transactions in a separate collection? /cc @ldmberman 


> Thinking about corner cases (capital letters represent nodes):
>
>    Genesis with A, B, C, D, E
>    A pushes a VALIDATOR-ELECTION e₁ to remove E (upsert-validator with power 0)
>    E pushes a VALIDATOR-ELECTION e₂ to add F
>    E transfers its vote for e₂ to Alice
>    A, B, C, D vote positively for e₁, and E gets kick out of the network

> According to the criteria discussed, e₂ is still valid. Why not?

I have mentioned this to @ldmberman during our previous discussions and would like to raise my concern again. The below formulation,

> The following has to be true in order for the ith election to be concluded (we check it on every DeliverTx and EndBlock):
> ```
> Pi + Pi-1 + ... + P0 < 1/3T
> Vi > 2/3 (T + Pi-1 + ... + P0)
> ```

allows for a very generalized set of validator set state mutations. We don't have an exhaustive study of what validator set transitions can occur and their corresponding rational correctness (though they are logically valid according to the above criteria).

kansi:
After thinking about this for a while and looking at how [Casper FFG](https://arxiv.org/pdf/1710.09437.pdf) allows dynamic validator set, I have tried to consolidate my thoughts,

### What is the scope of this feature?
We wish to allow the network to add/update/remove validator set dynamically. For which this BEP proposes an automated and synchronized way of achieving it.

### Security
Since the solution involves triggering an action via the means of a valid transaction it implies that the action will be triggered everytime the transaction is (re)played. So the attack vectors will be around how to get an election transaction to be considered valid even when it's not supposed to be a valid election (rationally or logically).

### Current implementation
The [current implementation](https://github.com/bigchaindb/bigchaindb/pull/2314) adopts the following logic,
- Store a snapshot of the current network validator set in the election transaction.
- Verify if the `>2/3` of the votes have been casted and check if the current network validator set matches to the snapshot in the election transaction. If it doesn't then the election can never be concluded.

The above logic is quite constrained and will allow for only one election transaction to be concluded i.e. consider `(V1, t1)`, `(V2, t2)` be the validator sets at time `t1` and `t2` such that `V1 -> V2`. Then there is only one election `Ei` proposed between `t1 <= Ei < t2` that can be concluded and rest will be rejected. One property that is ensured by the above constraints is that the validators present at the time of proposing an election are always present at the time of election conclusion. Hence, there can be no *rational* argument for declaring a validator set mutation invalid.


### Generalizing constraints
In order to generalize the above contraints and allow multiple transactions to be concluded the following new contraints have been suggested,

> ```
> Pi + Pi-1 + ... + P0 < 1/3T
> Vi > 2/3 (T + Pi-1 + ... + P0)
> ```

The above constraints ensure that any election can be concluded if the current super-majority supports the change. Needless to say, the total allocated votes during election proposal need to form the current super-majority for the election to be concluded. It is not really clear whether all the validator set mutations according to the above constraints can be considered rational.


In an attempt to quantify what mutations are rational I propose the following definition,


**Definition**: A change is rational if the present and the future agrees upon it.


From the above definition, we deduce that validator set mutation needs a super-majority not only from the current validator `V_current` set but from the validator set that occurs after the current change is applied i.e. `V_current+1`.


Rethinking this below scenario,

> Thinking about corner cases (capital letters represent nodes):
>
> -  Genesis with A, B, C, D, E
> -  A pushes a VALIDATOR-ELECTION e₁ to remove E (upsert-validator with power 0)
> -  E pushes a VALIDATOR-ELECTION e₂ to add F
> -  E transfers its vote for e₂ to Alice
> -  A, B, C, D vote positively for e₁, and E gets kick out of the network

Recall that even though `E` is out of the validator set it can still cast a vote for `e₁` which would be considered valid. So only 2 votes are required from the current validator set to achieve supermajority (i.e. `A`, `B`, `E`). In this situation even though the current supermajority didn't agree, the election would be concluded anyways.

Now, if we adopt the above-stated definition, then the election is concluded iff the current and future supermajority also agrees i.e. current supermajority requires `3` votes but the future supermajority requires `4` votes. So `(A, B, E, C)` is required to conclude the election.


Rationally speaking, the future supermajority cannot complain about the actions of past supermajority when they have been taken in to account.

@ldmberman  @vrde  Let me know your thoughts on this.

ldmberman:
> Recall that even though E is out of the validator set it can still cast a vote for e₁ which would be considered valid. So only 2 votes are required from the current validator set to achieve supermajority (i.e. A, B, E).

This is not true. We [defined](https://github.com/bigchaindb/BEPs/pull/45#issuecomment-398825857) `Vi` as:

> Vi - the total power of validators voted on the ith election.

Therefore, after `E`'s power turns into a pumpkin, its vote does not take effect.

In general, we do follow the "the current supermajority concludes the change" and "each individual change can be no bigger than 1/3 of the current power" rules.

kansi:
@ldmberman 
> Therefore, after E's power turns into a pumpkin, its vote does not take effect.

I don't get this point. If `E` initiated `e₂` and then voted on it but later gets kicked out then how does its votes get neglected. That fact the `E` is not in the current validator set doesn't matter because the we are couting the sum of output `amounts`.

ldmberman:
> That fact the E is not in the current validator set doesn't matter because we are couting the sum of output amounts

But we can make it matter, do we? That's another point we need to thoroughly document, to avoid the future confusion.

By ` the total power of validators voted on the ith election` I have been assuming the current total power of the validators that voted. That is also why it has an index - `i`, cause the power may change after each election and `Vi` not necessarily equals to `Vi-1`.

kansi:
> By the `total power of validators voted on the ith election` I have been assuming the current total power of the validators that voted. That is also why it has an index - `i`, cause the power may change after each election and Vi not necessarily equals to `Vi-1`.

What is the point of assigning votes in the `outputs` when creating an election?

> But we can make it matter

Sure, but my question is how many such conditions are necessary i.e. what is the total set of constraints necessary for an election?

ldmberman:
> What is the point of assigning votes in the outputs when creating an election?

Output amounts can be used to process vote delegation. For instance, as a voter with power=40 you can delegate your vote to 2 other people each with power=20. Over time each of them may get more power but they can only put at most 20 voting power into that particular election.

To recap on the constraints we have:

Checked when the election transaction is **created**:

- The transaction conforms to the schema.
- Every single output amount MUST match the power of the validator it has been transferred to.
- The amount of the transaction MUST be the amount of total power of the validators.

I even think we can relax these two but we can do it in the future - relaxing is backwards-compatible.

Checked when the election transaction is **concluded**.

- The amount MUST be spent fully in the outputs of the VALIDATOR_ELECTION transaction.
- The amount spent by every validator MUST be less than or equal to the current power of this validator.
- The sum of amounts is bigger than 2/3 of the current total validator power.
- The power change is smaller than 1/2 of the current total validator power.

Does it make sense to you?

ldmberman:
> What I am asking is, are you suggesting to consider each validators current power as their number of votes casted on an election which was created in past (assuming in past that particular validator didn't have the current power)?

No, the sum of amounts is considered to be the number of votes as in `The sum of amounts is bigger than 2/3 of the current total validator power.`.

But for each validator the amount must not be bigger than its power. It is not always true since the power of the validator may decrease as in the example with `E` which gets kicked off.

> Why isn't achieving >2/3 votes on the election enough? Why MUST all the votes be spent?

@vrde suggested it [here](https://github.com/bigchaindb/BEPs/pull/45#issuecomment-400309747) but I think we should relax this constraint.

Sorry `1/2` was a typo - off course, it's `1/3` there.

ldmberman:
> Sometime later in the future, the validator set becomes (M,N,O,P) again with total power 40 at this point A,B,C,D decide to vote and add F. Does seem like a valid election. IMHO No,

It won't be a valid election since the first conclusion requirement is broken:

> The amount spent by every validator MUST be less than or equal to the current power of this validator.


> Furthermore, we don't have a proper way of terminating election conclusion i.e. every new vote after >2/3 votes have been achieved will cause the conclusion logic to be triggered which might lead to undesired behaviour.

If while building a block you see that `> 2/3` of votes was there before the block, you do not process the conclusion.

ldmberman:
If we decide to always wait for all the election amounts to be spent, then how should validators express the desire to not include a new validator?

ldmberman:
Also, the system has to work as long as up to 1/3 of machines fail in arbitrary ways. If we require all the amounts to be spent that effectively means that malicious nodes can stall the elections.

21/README.md
```diff
@@ -0,0 +1,117 @@
...
+## Technical details
+
+To solve the aforementioned issue, following change to the behaviour of `upsert-validator` is being proposed. We use the transaction election process (TEP) proposed in [BEP 18](../18) to automate and synchronize the operation on the validator set.
+
+Consider a network of 4 nodes `{A,B,C,D}`. If a node `A` wishes to add a new node `E` to the network then following steps should be followed,
+
+1. Node `A` executes
+
+  ```
+  $ bigchaindb upsert-validator E_PUBKEY E_POWER E_NODE_ID --private-key /home/user/.tendermint/config/priv_validator.json
```
vrde: The new command `upsert-validator` has two subcommands:
- `show-request`
- `approve-request`

If no subcommand is given, then the default behavior is to start a new election. Should we have a specific command for that, something like `create-request`?

PR #47, opened 2018-05-29
Problem: BEP-10 misses performance insights.

ldmberman:
Solution: Document the profiling and benchmarking made.


kansi:
Thanks for giving a detailed insight :clap: . 

I have a concern though i.e. calling python from golang. I looked at the provided [link](https://www.datadoghq.com/blog/engineering/cgo-and-python/) and it seems like a decent amount of work. Any idea on how much time it might take to implement that approach?

ldmberman:
> I have a concern though i.e. calling python from golang.

This is unrelated; just the same "Rationale" section.

vrde:
> I can definitely move the details to a separate markdown file linking it from README.md - is it what you mean?

Yes exactly.

> 
>     @z-bowen and ran some other benchmarks
> 
> Is it the comparison of a samle in-process and over-the-wire ABCI proxy apps? We need to include it here somehow too!

Yup, I was thinking that we can put in the `README.md` what we discovered from the tests we ran, and in two different appendixes (can be two other markdown files) we put all the details.

PR #52, opened 2018-06-19
Problem: Tendermint breaks under sustained load

z-bowen:
Solution: Created BEP-19 (informational) to detail my findings about this issue. Also included a proposed strategy to mitigate the problem.


18/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+First, we can use the sentinel to protect against DoS attacks, by limiting the number of requests from any given IP address to a reasonable number.
+
+                                            _____________________                                ____________
+    User_1, 100,000 requests/s------------>| Sentinel            |   User_1, 995 requests/s     |            |
+    User_2, 1,000 requests/s-------------->| max_req=2,000/s     |---User_2, 995 requests/s---> | Tendermint |
+    User_3, 10 requests/s----------------->| pause_interval=100s |   User_3, 10 requests/s      |____________|
+                                           |_____________________|   Break every 100s
+
+Second, we can use the sentinel to break and reinitialize connection with Tendermint as a mitigation for the perfomance degradation.
```
ldmberman: > to break and reinitialize connection

You mean, to control the rate at which transactions are posted? There is no need in breaking connections.

18/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+This could be implemented in two ways. First, we could simply experiment to find the optimal frequency with which we hang up and call back to Tendermint. This has the advantage of being easy to engineer. However, we could also implement active monitoring from the sentinel layer, where we measure the transaction rate and hang up if it falls below a certain threshold. This offers the advantage of being a more general solution that will respond to changes in performance as different version of Tendermint are rolled out, or when the system is run on different hardware, for example.
```
ldmberman: Might make sense to mention [AIMD](https://en.wikipedia.org/wiki/Additive_increase/multiplicative_decrease) here.

18/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Test System
+All tests were run on hardware with the following specifications
+
+* Processor: Intel(R) Xeon(R) CPU E5-2650 v3 @ 2.30GHz (10 cores)
+* Memory: 15GB
+* OS: Ubuntu 18.04 LTS
+* Tendermint version: 0.20.0-27bd1dea
+
+## Tests
```
ldmberman: Did you test with `create_empty_blocks=false`?

May be also worth trying to run with `recheck=false`. A blind guess - a rate would go up higher but the general degradation pattern will stay.

ldmberman:
Great work, thank you!

18/README.md
z-bowen: That was the conclusion I reached from this study. Breaking the connection to Tendermint every 100s or so mitigates the performance degradation. So if we have the sentinel, not only are we protected from DoS attacks, but we have a single point where we can manage this connection refresh.

18/README.md
z-bowen: I don't think rate limiting the txs will help. I see the same performance decay even when I run at 10 requests per second.

18/README.md
z-bowen: I tried it with `create_empty_blocks=false` and `true`. It didn't make a difference, since while the test is running, there is a constant stream of requests, and hence no empty blocks get minted.

18/README.md
ldmberman: I am not talking about rate limiting. On the contrary, we can allegedly maintain a high external rate limit by varying the pressure we put internally on Tendermint. For that, we would need a buffer in between.

18/README.md
ldmberman: I mean, you do not have to break the connection to put the pressure off the system - a more efficient approach is to simply stop sending transactions.

Technically though, in order to do it, you would need to patch the `tm-bench` tool or create a custom one.

We do not have to do it now, but I would still use the term "pause the pressure" (or a similar one) instead of "break and reinitialize the connection".

z-bowen: Fair point. I'll update the language to indicate that this hasn't been tested.

18/README.md
z-bowen: Yes. My idea here was that the sentinel server would remain open to external requests all the time, and store excess requests in a buffer. It would manage the forwarding of requests to TM without breaking connections with the outside world.

18/README.md
z-bowen: Added a paragraph to make this explicit.

ldmberman: Great, thanks!

18/README.md
```diff
@@ -0,0 +1,128 @@
+```
+shortname: 18/tm_stress
```
kansi: BEP-18 is already taken by a previous PR https://github.com/bigchaindb/BEPs/pull/44

kansi: I think BEP-20 is available as BEP-19 is taken too.

19/README.md
```diff
@@ -0,0 +1,128 @@
+```
+shortname: 19/tm_stress
```
kansi: [BEP-19](https://github.com/bigchaindb/BEPs/pull/45) is taken too. I as I [previuosly](https://github.com/bigchaindb/BEPs/pull/52#discussion_r196709563) you can use BEP-20

PR #53, opened 2018-06-21
Problem: BEP-14 lacks timeout handling and exp backoff.

ldmberman:
Solution: Refine the algorithm to incorporate timeout and exp backoff. Explicitly describe how common and custom headers are set together.

### DIFF
--- a/14/README.md
+++ b/14/README.md
@@ -4,6 +4,7 @@ name: Guidelines to Improve Drivers Reliability
 type: standard
 status: raw
 editor: Alberto Granzotto <alberto@bigchaindb.com>
+contributors: Lev Berman <ldmberman@gmail.com>
 ```
 
 # Guidelines to Improve Drivers Reliability
@@ -16,7 +17,7 @@ A BigchainDB driver allows developers to integrate their applications with a Big
 
 Assumptions on BigchainDB availability and correctness are bound to the concept of **network of nodes**. If a driver connects to only one node, the reliability of BigchainDB perceived by the end user is harmed.
 
-This BEP aims to improve the reliability of the drivers, and does **not** address how drivers should implement a [light client protocol][tm:light-client].
+This BEP aims to improve the reliability of the drivers, and does **not** address how drivers should implement a [light client protocol][tm:light-client]. In other words, all the nodes the driver communicates to have to be trusted.
 
 ## Specification
 First, let's take a look on how the current implementation works. In order to have a usable instance of a driver, the developer needs to specify the URL of the BigchainDB node to connect to.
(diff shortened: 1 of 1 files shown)


14/README.md
```diff
@@ -134,28 +136,34 @@ For this BEP, we analyze a round-robin connection strategy, similar to [Round-ro
 In this scenario, every request is forwarded to a different node, distributing the load to all the nodes, and allowing recovering from connection errors by forwarding the failed request to another healthy node.
 
 This strategy needs to keep track of the following values:
-- `E`, the list of BigchainDB endpoints passed at initialization time (we use the Python notation `len(L)` to get the length of the list).
-- `i`, the index of the BigchainDB node to use for a request (`0 ≤ i < len(L)`).
-- `T`, a list of timestamps. It has the same length of `L`. On initialization, all elements contain `0`. It is used to track the availability of the nodes. We will see in a second how it is used.
-- `DELAY`, in case a node is unreachable, how much milliseconds we should wait. This is used in combination with `T`.
+
+- `E`, the list of BigchainDB endpoints passed at initialization time.
+- `timeout`, user-specified timeout, by default there is no timeout.
```
vrde: If a developer initializes the driver with a wrong endpoint, the driver will forever fail the request. Not having a default timeout might make the code hang forever.

I'd put a default timeout (just like any socket connection) of [20 seconds](http://www.sekuda.com/overriding_the_default_linux_kernel_20_second_tcp_socket_connect_timeout), that can be initialized to `None` if the user wants to **not** have a timeout.

14/README.md
```diff
@@ -134,28 +136,34 @@ For this BEP, we analyze a round-robin connection strategy, similar to [Round-ro
...
+- `T`, a list of timestamps. It has the same length as `E`. On initialization, all elements contain the minimal possible timestamp `datetime.min`. It is used to track the availability of the nodes.
```
vrde: `datetime.min` is Python specific. Instead of using a Python object, milliseconds since epoch can be used, and the `min` can be just `0`.

14/README.md
```diff
@@ -134,28 +136,34 @@ For this BEP, we analyze a round-robin connection strategy, similar to [Round-ro
 In this scenario, every request is forwarded to a different node, distributing the load to all the nodes, and allowing recovering from connection errors by forwarding the failed request to another healthy node.
 
 This strategy needs to keep track of the following values:
-- `E`, the list of BigchainDB endpoints passed at initialization time (we use the Python notation `len(L)` to get the length of the list).
-- `i`, the index of the BigchainDB node to use for a request (`0 ≤ i < len(L)`).
```
vrde: `i` is used in the definition of the algorithm below, but it is now removed. I'd restore this line.

14/README.md
```diff
@@ -134,28 +136,34 @@ For this BEP, we analyze a round-robin connection strategy, similar to [Round-ro
...
-    1. return the result or the error
-1. if `current_time_ms ≤ T[i]`:
-    1. move the index to the next node `i = (i + 1) mod len(N)`
-    1. return error
+1. pick the node `i` with the minimal `T[i]`, that is the earliest available node
+1. wait until `current_time_ms > T[i]` or timeout is expired
+1. if timeout is expired, return
+1. if timeout is not expired, make the request
+1. if timeout is expired while the request is in progress, return
+1. in case of a successful response, set `retry_count` to 0, set `T` to `datetime.min`, return
```
vrde: If we set `T` to `0`, then we end up hitting always the same "successful" server. By setting `T` to the current time in milliseconds, we can achieve something similar to a round robin. This helps distributing the workload to different servers. What do you think?

ldmberman: The successfull node has bigger probability to further be successful so by exploiting the same node we optimize the processing time. On a side note, I think this has the direct relation to the [multi-armed bandit problem](https://en.wikipedia.org/wiki/Multi-armed_bandit).

Every driver user assembles the endpoint list by himself so the workload distribution should be close to uniform under the assumption that driver users do not have common preferences for nodes.

In case you want to speed up the upload of a bunch of transactions through parallelization, the round-robin inside the driver won't help to achieve that. Currently, you can create multiple driver instances each with a different node list and do requests in parallel. Aditionally, in the future we may accept an endpoint as a parameter so that users can do parallel requests to multiple nodes using a single driver instance.

vrde: Fair enough, so the round robin is applied on failure, not on success.

14/README.md
```diff
@@ -134,28 +136,35 @@ For this BEP, we analyze a round-robin connection strategy, similar to [Round-ro
...
```
vrde: Should `datetime.min` be `0`?

ldmberman: Yes! Forgot to update it here, good catch.

PR #56, opened 2018-06-28
Problem: No way to handle new transaction models and storage schemas (BEP-42)

kansi:
Solution: Propose required strategies


42/README.md
```diff
@@ -0,0 +1,71 @@
+```
+shortname: [42/HARD-FORKS]
+name: Handling Hard Forks
```
ldmberman: I would call it `Handling Forks` because the `Type 1` approach below can be used to govern "Soft Forks" (backwards-compatible validation logic updates) too.

42/README.md
```diff
@@ -0,0 +1,71 @@
...
+    - Change in hashing functions
+    - Change in schema
+2. Change in storage schema
+
+## Handling Hard Forks
+
+### Type 1 scenarios
+In case of change to spec it might not be possible to port all existing validated transaction to the new format. In such cases one needs to make a distinction between transactions which have already been validated and stored in the Blockchain from new incoming transactions. Furthermore, a suitable trigger is needed so that the network can collectively jump to the new logic else there could be situations wherein the network might not be able to commit any new blocks because of honest nodes running different validation logic. So, it is apparent that a mechanism for transitioning to this new validation logic is need. Below is one such procedure to achieve the same,
+
+- In order to decide when the validation logic should be upgraded to its newer version a TEP is initiated which defines the future block `height` at which to upgrade and the corresponding `BigchainDB` class to use.
```
ldmberman: Just some thoughts: it would be cool to be able to fork off of a height from the past as well. For example, as soon as an election with the height from the past is concluded, the app abandons the corresponding part of the chain and starts to build up from this height again. But I am afraid it can be done only in sync with Tendermint so there should be a similar feature there, right?

42/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+   NOTE: It is assumed that the new validation logic is implemented by extending the existing `BigchainDB` class and at the time of starting BigchainDB an instance of each of these classes is passed to the ABCI `App` i.e.,
+   ```python
+   validation = {
+       "v1": {
+           "start_block_height": 0,
+           "validation_class": "BigchainDB",
+       }
+       "v2": {
```
ldmberman: Over time, the versions may accumulate so it might be a good idea to design a deprecation process offering a way to archive the old blocks and strip the old code off. We might not want to include it here to keep it simple though.

42/README.md
codegeschrei: can you mention what TEP stands for? Maybe one of those: https://www.acronymfinder.com/TEP.html :D
I like `Tender Evaluation Plan` or `Tactical Earth Penetrator` 💅

42/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Type 2 scenarios
+
+These scenarios don't actually impact the sanctity of the blockchain but rather the shape of the data. In order to handle such scenarios following two approaches can used,
+
+#### Approach 1
+
+Create migration scripts for each new upgrade i.e. with each new release which results in a change in the shape of the data we create corresponding migrations scripts. Needless to say that writing such scripts might be tedious and in the worst case users can only incrementally upgrade to newer version
+
+#### Approach 2
```
muawiakh: This sounds like a decent approach. but some sort of basic upgrade prototype should be made before, we merge this BEP. I am quite unsure about how Tendermint might behave at run time changes for validator/followers in `config.toml`

kansi: > but some sort of basic upgrade prototype should be made before

Could you please elaborate?

> I am quite unsure about how Tendermint might behave at run time changes for validator/followers in `config.toml`

The running validator's `config.toml` is not changed, rather we create a follower node which we configure to talk to this validator node. Once the follower node is up to pace with the validator node we change the `proxy_app` of the validator node to the BigcahinDB server connected the follower node.

42/README.md
kansi: This would be a tricky problem i.e. whenever a replay of all the blocks is required the old code will be needed. I am happy to explore possible these ideas further but that discussion would be out of the scope of this BEP.

42/README.md
ldmberman: So, a theoretical Tendermint feature may allow to avoid replaying this tail. But you are right - the initial step towards the solution can be done with what we've got already!

> allowing for such a behaviour would be an anti-pattern

If there is a good reason for such a fork, people will make it anyway but they would need to orchestrate it themselves. Why do you think envisaging the tooling for that is an antipattern?

kansi: The idea you are a suggestion here stems from the desire to delete data from a blockchain. Since blockchains are _viewed as immutable datastores_, such a feature would seem anti-pattern. An argument in favour of such a feature comes about when we tweak the definition of blockchain a little i.e. software to reach consensus in presence of byzantine actors (which is what they actually are). So, if the network agrees to undo `m` blocks it should be able to do so.

All in all what I wish to state is that your suggestion is aiming to solve a subset or a special case of a general problem. If the users ask for a solution for this specific use case then we can implement such a feature but until then we can let it be.

42/README.md
```diff
@@ -0,0 +1,71 @@
...
+1. Change in any of the transaction specs
+    - Change in signature algorithm
+    - Change in hashing functions
+    - Change in schema
+2. Change in storage schema
+
+## Handling Hard Forks
+
+### Type 1 scenarios
+In case of change to spec it might not be possible to port all existing validated transaction to the new format. In such cases one needs to make a distinction between transactions which have already been validated and stored in the Blockchain from new incoming transactions. Furthermore, a suitable trigger is needed so that the network can collectively jump to the new logic else there could be situations wherein the network might not be able to commit any new blocks because of honest nodes running different validation logic. So, it is apparent that a mechanism for transitioning to this new validation logic is needed. Below is one such procedure to achieve the same,
```
ttmc: My thinking about the "Type 1 scenario" was that if a new transaction spec (e.g. v3) gets implemented, then nodes can accept new v2 and v3 transactions for a while. Before a node upgrades its BigchainDB Server software to a version that recognizes v3 transactions, it will just think that v3 transactions are invalid, so for a while, v3 transactions will be considered invalid by the network. Once > 2/3 of the nodes upgrade, then v3 transactions will have a chance of being considered valid. No TEP or vote is needed.

Later, the nodes could upgrade their BigchainDB Server software to software that rejects _new_ v2 transactions, but can still validate old v2 transactions.

This process can be repeated for v4, v5, etc.

42/README.md
```diff
@@ -0,0 +1,71 @@
+```
+shortname: [42/HARD-FORKS]
+name: Handling Forks
```
ttmc: The name of this BEP forces a particular solution strategy rather than identifying a problem, surveying possible solutions, and recommending a particular solution. There are two problems, really, which this BEP currently calls Type 1 and Type 2. It's probably okay to include both in one BEP. Maybe the BEP title should be something like "Handling new transaction models and storage schemas".

I'm not sure that the proposed solutions (as I understand them) are correctly described as "hard forks." When I picture a blockchain hard fork, I imagine going back to a previous block and rewriting the chain from there, according to new rules. The old chain and the new chain can be pictured like a [tuning fork](https://en.wikipedia.org/wiki/Tuning_fork).

42/README.md
ldmberman: @ttmc we need to distinguish old transactions from new transactions. Upon replay, how does a validator know that he needs to switch to `v2` at height `5` and then to `v3` at height `20`? The approach described here basically leverages the election process to address these problems.

42/README.md
ldmberman: The difference between soft and hard forks is determined by [backwards-compatiblity - it's not about chain rewrites](https://en.wikipedia.org/wiki/Fork_(blockchain)#Hard_fork).

Approach 1 here can be used to handle both soft and hard forks.

`Handling new transaction models and storage schemas` sounds like a good name to me.

42/README.md
ttmc: A transaction contains a field/key "version" that says which transaction spec version it follows. Currently, the only allowed value is "2.0". The node will validate the transaction using the validation rules for that transaction version number.

ttmc: I was thinking about handling changes to the transaction spec and realized that there are at least two cases:

1a) There are no problems with the old transactions, but a new transaction spec has been devised and it is better in some ways, so people would like to use transactions following the new spec. This case can be handled as I described in my comment above, without any need for proposals, elections or changing old blocks.

1b) Some problem is found with old transactions, such as a security flaw. The course of action will depend on the nature of the problem with the old transactions, but it will have to be more drastic. For sure, new transactions following the flawed spec shouldn't be allowed, and so the proposal & election process can be used to trigger a change to a new/fixed spec. It might be necessary to do more, e.g. to prevent unspent outputs from being transferred from old/broken outputs to thieves. It all depends on what the problem is.

ldmberman: > This case can be handled as I described in my comment above, without any need for proposals, elections or changing old blocks.

Yes, you are right. Also, during replay the versions of the transactions are known so we do not strictly need the election mechanism for the case when the new version can not validate old transactions.

On the other hand, the election process can make the migration smoother. Without it, how do you know if you can post new transactions already? You can run trial and error but it is costly.

kansi: @ttmc The case I am tried to address here is 1b i.e. we might not need to do anything special in case there is new transaction spec which facilitates a new behavior (or a backwards compatible change). The main concern this strategy addresses is when the validation rules for an existing transaction have to be upgraded such that the new rules are incompatible with the older rules. The reason for doing so might be a security flaw of something else. So the general aim to ensure that no new transactions that invalid according to new validation rules are committed to the blockchain. But we also need to ensure that the older transaction if and when replayed are indeed validated. So it seems that we have an explicit declaration which indicates the activation of new rules across the network.

> It might be necessary to do more, e.g. to prevent unspent outputs from being transferred from old/broken outputs to thieves. It all depends on what the problem is.

True, depending on the scenario there would be other required changes but these changes would be a part of the new validation logic.

ttmc:
The Tendermint team is discussing adding some features to Tendermint to make it easier for app developers (such as BigchainDB) to update to new app versions. See:

https://github.com/tendermint/tendermint/pull/1983/files

42/README.md
ttmc: It seems that Tendermint is very strict: you should never have a time interval (sequence of blocks) where some validators think a transaction is valid and others think it's invalid: they'll consider each other to be Byzantine validators! That must be avoided, so an election (TEP) will be needed to get the validators to agree on a future block height where new rules will come into effect.

42/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Following describes how to archive and start a new blockchain,
+- Create a TEP which proposes a block height `h` at which the current blockchain will stop creating any new blocks.
+- Once the TEP concludes and the height `h` has been chosen, validators stop proposing new blocks once height `h` has been reached.
+- At height `h` the merkle root of the UTXO of the current blockchain is provided as the genesis state of the new blockchain.
+- The validator set at height `h` should be used as the validator set in the genesis of the new blockchain.
+- Since from Tendermint's point of view this is a new blockchain, block height will start from `0` which implies that BigchainDB would need to achieve the existing blocks collection. All other collections would remain as is.
```
kansi: NOTE: It is not confirmed if Tendermint will acknowledge the new blockchain as a fork and start from height `h` rather than height `0`. In case Tendermint doesn't acknowledge then BigchainDB will have to maintain the block heights such that forking at height `h` is accounted for.

42/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Following describes how to archive and start a new blockchain,
+- Create a TEP which proposes a block height `h` at which the current blockchain will stop creating any new blocks.
+- Once the TEP concludes and the height `h` has been chosen, validators stop proposing new blocks once height `h` has been reached.
+- At height `h` the merkle root of the UTXO of the current blockchain is provided as the genesis state of the new blockchain.
+- The validator set at height `h` should be used as the validator set in the genesis of the new blockchain.
+- Since from Tendermint's point of view this is a new blockchain, block height will start from `0` which implies that BigchainDB would need to archive the existing blocks collection. All other collections would remain as is.
```
kansi: NOTE: It is not confirmed if Tendermint will acknowledge the new blockchain as a fork and start from height h rather than height 0. In case Tendermint doesn't acknowledge then BigchainDB will have to maintain the block heights such that forking at height h is accounted for.

PR #66, opened 2018-08-09
Problem: Tx specs allowed TRANSFER asset.id == null

ttmc:
Solution: Edit the tx specs to only allow TRANSFER asset.id == {'id': <64-char hex string>}


ttmc:
Note: The JSON schema files for TRANSFER transactions are already only allowing the asset.id to be an "#/definitions/sha3_hexdigest"

PR #68, opened 2018-08-14
Problem: Tendermint chain migration plan is incomplete.

ldmberman:
Solution: Draft a migration plan. Address the problem of switching Tendermint chains. Describe how to replay the chain after a migration. Mention the implications on HTTP API.


42/README.md
```diff
@@ -70,14 +71,129 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+Validators have to install and launch the new version of Tendermint. They need to prepare a new `genesis.json`. The new genesis file has to contain the validator set, the application hash (both at the migration height), and the identifier of the new Tendermint chain (`chain_id`).
+
+The application hash is calculated by computing the Merkle tree root of the current UTXO.
+
+`chain_id` is generated and stored by BigchainDB upon conclusion of the migration election. When Tendermint sends this ID as part of the `InitChain` ABCI request, BigchainDB understands that the user has switched to a new Tendermint version so BigchainDB switches to accept new transactions and blocks.
+
+To offer a convenient way to get the data, we propose a CLI command:
+
+```
+$ bigchaindb status
```
kansi: The command seems very general i.e. just by looking at the command is not clear that it would return the migration data. I would prefer a different name i.e. something like `bighciandb migration get-genesis <migration-id>`?

42/README.md
```diff
@@ -70,14 +71,129 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+```
+$ bigchaindb migration status <migration-id>
+```
+
+Outputs:
+```
+votes_recieved=<Sum_of_votes_recieved>
+votes_allocated=<Sum_of_votes_allocated_in_election>
+network_size=<Total_network_power>
+```
```
kansi: I think it is possible to definitely state that the migration concluded or not. We can just state if the migration has concluded or not.

ldmberman: Makes sense. Here I wanted it to look similar to how it looks for `upsert-validator`. We can add an additional `migration_concluded=` line for both, what do you think?

42/README.md
ldmberman: My thinking is `bigchaindb status` can be as well useful outside of the migration scope. For instance, it is a good way for the operator to know the current validator set.

I am also thinking to change `bigchaindb migration dump/apply` to `bigchaindb dump/apply`. This way, node operators can back things up every once in a while so that if their storage experiences a critical failure, they can replay the chain from the dump. In the future, we may include Tendermint's `genesis.json` there (for instance, after we package a single binary).

Does it sound good?

42/README.md
```diff
@@ -70,14 +71,129 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+#### 2. Replay the blockchain
+
+In order to replay the chain up to the migration height, we need to either keep the old Tendermint version running or skip replaying the corresponding part of the chain = archive the chain. We consider the former to be a huge hassle so we conceive an archiving scheme further in this chapter.
```
ttmc: The last sentence ends ambiguously with, "...so we conceive an archiving scheme further in this chapter." Conceive? Chapter?

Maybe, "...so we describe an archiving scheme elsewhere in this BEP."

42/README.md
```diff
@@ -70,14 +71,129 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+#### 3. Start the new chain
+
+Validators have to install and launch the new version of Tendermint. They need to prepare a new `genesis.json`. The new genesis file has to contain the validator set, the application hash (both at the migration height), and the identifier of the new Tendermint chain (`chain_id`).
+
+The application hash is calculated by computing the Merkle tree root of the current UTXO.
```
ttmc: That's not how the application hash is calculated today, but I guess that's an idea for how to calculate it in the future?

Maybe the details of _how_ to compute the application hash should be left out of this BEP, because that's one of the things that can change over time. Vanshdeep started another BEP (in a pull request) to specify how to compute the application hash, in the future.

42/README.md
```diff
@@ -70,14 +71,129 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+}
+```
+
+Validators upgrading Tendermint are supposed to execute the command above and copy the validators, the app hash, and the chain ID into their `genesis.json`.
+
+Validators joining the network after the migration take the genesis file from existing validators, as usual.
+
+#### 4. Join the network after a migration
+
+When a validator joins the network after a migration, it does not suffice for him to know `genesis.json` - he also needs the archive of the old chain (heights from 0 to the latest migration height).
```
ttmc: There might be more than one migration, so the initial chain might go from height 0 to 33277, the second chain might go to height 88234, the third chain...

42/README.md
ldmberman: This way of computing hash was described here prior to this patch. We do need to know how to compute the hash to be able to migrate.

42/README.md
kansi: Yes :+1:

42/README.md
```diff
@@ -70,14 +71,129 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+Additionally, BigchainDB HTTP API usage has to be seamless - all the HTTP responses has to look like there were no migrations.
+
+Below we describe how to perform a migration while meeting the postulated requirements. Note that there is no limitation on the number of consequent migrations that can be performed this way.
+
+#### 1. Stop building blocks
+
+A CLI command can be used to instruct the node to stop building blocks at a chosen height.
+
+To make sure no blocks are committed after the agreed height, we propose an election process:
```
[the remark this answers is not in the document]
kansi: Following up on that, when an election concludes before the proposed migration height `h` and there are no new transactions in the system such that blockchain can reach height `h` then what is supposed to happen?

42/README.md
kansi: I would prefer a separation of concern i.e. one command only does one thing. If we choose `bigchaindb status` to list the genesis of the next migration then it can only do that nothing else.

42/README.md
ttmc: You definitely need to know how to compute the application hash to migrate, but the method of computing it can change from one chain to the next.

42/README.md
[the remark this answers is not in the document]
ldmberman: Good points! I will document these cases.

42/README.md
ldmberman: The idea is to make `bigchaindb status` show the current state of the system, similarly to what the Tendermint's RPC endpoint `/status` is doing.

Figuring out the genesis of the migrated chain is still a separate concern which would in this case involve using the generic status command.

Basically, my point is it might be inconvenient to create a dedicated command for every small concern. Imagine the amount of commands we might have - `bigchaindb get-validators`, `bigchaindb get-chain-id`, `bigchaindb migration get-validators`, `bigchaindb some-integration get-validators`, etc.

On a side note, it's not a good experience to copy and paste the data. Hopefully, in the near future we will have control over the Tendermint setup so that users can simply issue a `bigchaindb continue` type of command after downloading the new binary - it will then automatically pick up genesis data.

What do you think about `bigchaindb dump/apply` as a mean to replay the chain from a "checkpoint", useful even if there were no migrations?

42/README.md
ldmberman: This BEP should advise on a reasonable way to hash to use for the implementation. I think, the proposed way is more reasonable than the existing one. I will note it down that it might change.

42/README.md
ldmberman: Thinking about conclusion (also related to [this comment](https://github.com/bigchaindb/BEPs/pull/68#discussion_r210209288)), we need to display election status, one of `started`, `concluded`, `failed`.

Let's agree on the names for these statuses and document them here and in [BEP-21](https://github.com/bigchaindb/BEPs/tree/master/21).

42/README.md
ldmberman: Pondered about dumping a little more. I would keep `bigchaindb migration dump` as a command for dumping the state at the last migration height. A generic `bigchaindb dump` might still be introduced in the future but it will take a snapshot as of the last known height or at the given height, when specified.

Loading the dump is height-agnostic though so I would prefer a generic `bigchaindb apply <dump>` command.

Alternative names may be `bigchaindb migration create-snapshot`, `bigchaindb create-snapshot` (to be introduced later, not now), and `bigchaindb load-from-snapshot`.

What do you think?

42/README.md
kansi: Instead of `started` I would suggest `ongoing`? i.e. once an election transaction has been committed its status could be described as ongoing

ldmberman: Ok, I will add an `election_status=` line but there is no need to enumerate the statuses here. You can just do it in BEP-21.

42/README.md
ldmberman: Discussed offline - migration snapshots are only concerned with MongoDB so we can just use `mongodump`/`mongorestore` at this moment.

42/README.md
```diff
@@ -70,14 +71,129 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+- The initiator creates an election, which includes a block height.
```
shahbazn: Is the purpose of adding specific block height to add a kind of timeout for the election process? So that if the election concludes before or at the given block height the network halts and prepares to migrate or else the election is rendered obsolete?

42/README.md
```diff
@@ -70,14 +71,129 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+The command creates the `bigchaindb.dump` file that needs to be transferred to the new member. The dump contains all BigchainDB collections to restore the old chain from.
+
+The new validator needs to start BigchainDB and apply the dump:
+
+```
+$ bigchaindb migration apply <dump>
+```
+
+After the dump is applied, Tendermint can be started.
```
shahbazn: what happens if tendermint is started before the migration dump is applied? it will still sync with the current state of the new chain. I supposes tendermint can be started before or after the migration dump is applied but the bigchaindb server can only be started after the dump has been applied

42/README.md
ldmberman: I will update this part - there is no need to ask for a height there.

42/README.md
```diff
@@ -126,13 +126,13 @@ election_status=<Election_status>
...
 
-In order to replay the chain up to the migration height, we need to either keep the old Tendermint version running or skip replaying the corresponding part of the chain = archive the chain. We consider the former to be a huge hassle so we conceive an archiving scheme further in this chapter.
+In order to replay the chain up to the migration height, we need to either keep the old Tendermint version running or skip replaying the corresponding part of the chain = archive the chain. We consider the former to be a huge hassle so we describe an archiving scheme further in this BEP.
 
 #### 3. Start the new chain
 
 Validators have to install and launch the new version of Tendermint. They need to prepare a new `genesis.json`. The new genesis file has to contain the validator set, the application hash (both at the migration height), and the identifier of the new Tendermint chain (`chain_id`).
 
-The application hash is calculated by computing the Merkle tree root of the current UTXO.
+The application hash is calculated by computing the Merkle tree root of the current UTXO. Note that hash calculation is a subject to design in a separate BEP and can be changed in the future.
```
ttmc: I don't know understand why you kept a specific method of computing the application hash in this BEP.

Moreover, we don't calculate the application hash like that today, and I'm not convinced we'll compute it that way in the future. There's a lot more to the current BigchainDB state than the UTXO set. The full state includes all `asset.data` and `metadata` information, for example.

42/README.md
kansi: The BEP should specify the output of `bigchaindb status` command when there is no migration.

42/README.md
```diff
@@ -70,14 +71,133 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+Validators upgrading Tendermint are supposed to execute the command above and copy the validators, the app hash, and the chain ID into their `genesis.json`.
+
+Validators joining the network after the migration take the genesis file from existing validators, as usual.
+
+#### 4. Join the network after a migration
+
+When a validator joins the network after a migration, it does not suffice for him to know `genesis.json` - he also needs the archive of the old chain.
+There might be more than one migration, so the initial chain might go from height 0 to 33277, the second chain might go to height 88234, the third chain can go from 88234 up to the recent height. In this case the validator has to get an archive containing archived blocks from height 0 to height 88234.
+
+Such a validator needs to take a snapshot of the old chain from an existing member of the network. The snapshot can be created via a CLI command:
```
kansi: In the first iteration of this implementation, we can just recommend the user to use `mongodump` and `mongorestore` for achiving and loading data.

42/README.md
ldmberman: I used UTXO root since it can be used to do some useful verifications in the future, unlike the current hashing.

It's not indeed important at the moment because we will instruct the users to take the hash and the dump from the same validator - we do not anyway offer tools to verify the dump-hash integrity yet. So I can as well just recommend to use the already implemented way.

42/README.md
ldmberman: Yes, we need to be explicit about the importance of keeping/publishing the dump together with `genesis.json`.

ttmc: I think Tendermint stores more information than MongoDB stores, such as all the block headers. Shouldn't that be "dumped" as well?

42/README.md
```diff
@@ -70,14 +71,133 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+The new validator needs to start BigchainDB without Tendermint and apply the snapshot:
+
+```
+$ bigchaindb apply-snapshot <snapshot>
+```
+
+After the snapshot is applied, Tendermint can be started.
+
+Note that although the new node can join the network and work to some extent without applying the snapshot, it is not able to properly validate transactions so applying the snapshot is a must.
```
shahbazn: What effect will this have on the network? I suppose this node be treated as byzantine node?

ldmberman: Yes, it will. Without sufficient data, it may claim invalid transactions to be valid.

42/README.md
ldmberman: After a migration, Tendermint starts from scratch so no - there is nothing to dump.

ttmc:
I'm still a bit unclear about how the initial Tendermint state in a new chain gets set to the final state of the previous chain, but I suppose those details will get resolved once we start implementing (and we can revise this BEP to reflect what we do).

I will approve this PR.

ldmberman:
> I'm still a bit unclear about how the initial Tendermint state in a new chain gets set to the final state of the previous chain,

The initial Tendermint state is described by `genesis.json`. `chain_id`, the app hash, and the validator set is everything that matters. Does it help?

42/README.md
```diff
@@ -206,7 +206,18 @@ With every consequent migration, the migration height is overwritten.
 
 Since BigchainDB retains the blocks built by old Tendermint chains, the HTTP API offers the exact same experience as if there were no migrations.
 
+#### Migration election specs
+
+We introduce two transaction specs, with operation types [`TENDERMINT_MIGRATION_ELECTION`][spec_tendermint_migration_election] and [`TENDERMINT_MIGRATION_ELECTION_VOTE`][spec_tendermint_migration_election_vote], for the purpose of implementing migration elections. `TENDERMINT_MIGRATION_ELECTION` is an extension of `CREATE`; `TENDERMINT_MIGRATION_ELECTION_VOTE` is an extension of `TRANSFER`.
```
ttmc: I wouldn't call those two things "specs." They're two new operation types, to be added to the existing BigchainDB Transactions Spec v2 (BEP-13).

ttmc:
> The initial Tendermint state is described by `genesis.json`. `chain_id`, the app hash, and the validator set is everything that matters. Does it help?

Yes, thanks.

42/transaction_tendermint_migration_election.yaml
```diff
@@ -0,0 +1,28 @@
+---
+"$schema": "http://json-schema.org/draft-04/schema#"
+type: object
+title: Tendermint Migration Election Schema - Updgrade Tendermint after a breaking change in the chain
+required:
+- operation
+- asset
```
kansi: The `asset` field is not defined?

kansi:
Why is there is no transaction spec for creating a migration proposal?

ldmberman:
> Why is there is no transaction spec for creating a migration proposal?

It matches `CREATE`, except for the new operation.

kansi:
But then what is the value of the asset field?

ldmberman:
> But then what is the value of the asset field?

Follows the same constraints as in `CREATE` then.

42/README.md
```diff
@@ -70,15 +71,144 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
-- Create a TEP which proposes a block height `h` at which the current blockchain will stop creating any new blocks.
-- Once the TEP concludes and the height `h` has been chosen, validators stop proposing new blocks once height `h` has been reached.
-- At height `h` the merkle root of the UTXO of the current blockchain is provided as the genesis state of the new blockchain.
-- The validator set at height `h` should be used as the validator set in the genesis of the new blockchain.
-- Since from Tendermint's point of view this is a new blockchain, block height will start from `0` which implies that BigchainDB would need to archive the existing blocks collection. All other collections would remain as is.
+This section advises on how to approach backwards-incompatible upgrades of the Tendermint chain - the type of updgrades when Tendermint can not build blocks on top of the existing chain.
 
+BigchainDB operators need a convenient way to do the following:
+
+1. Stop building blocks when the old Tendermint chain is at a jointly chosen height.
```
ttmc: This line makes it seem like the operators have to agree on a migration height (e.g. 4592) ahead of time, but later on, you make it clear that the migration height is just whatever the current height happens to be when the election is concluded. Maybe change this line to read:

1. Stop building blocks for the old Tendermint chain at the same height as all the other operators.

42/README.md
```diff
@@ -70,15 +71,144 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+```
+$ bigchaindb election status <election-id>
+```
+
+Outputs:
+```
+votes_recieved=<Sum_of_votes_recieved>
+votes_allocated=<Sum_of_votes_allocated_in_election>
+network_size=<Total_network_power>
```
ttmc: Is this line correct? Network size and total network power don't seem like the same thing to me.

42/README.md
```diff
@@ -70,15 +71,144 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+#### 3. Replay the blockchain
+
+In order to replay the chain up to the migration height, we need to either keep the old Tendermint version running or skip replaying the corresponding part of the chain = archive the chain. We consider the former to be a huge hassle so we describe an archiving scheme further in this BEP.
+
+When a validator replays the chain after a migration, it does not suffice for him to know `genesis.json` - he also needs the archive of the old chain.
+
+Therefore, after the migration election is concluded, each validator has to create a chain archive and keep it together with `genesis.json`. If `genesis.json` is published somewhere, the archive should be published alongside.
+
+To generate the archive, validators can use the `mongodump` command (comes together with `mongod`):
```
ttmc: We now know that the blockchain stored by MongoDB and the blockchain stored by Tendermint are different because Tendermint also stores invalid transactions. Does this matter? Don't the invalid transactions go into calculating some block header hashes on the Tendermint side?

In other words, shouldn't we be archiving Tendermint's LevelDB? Maybe just archive both MongoDB and LevelDB?

(We're considering replacing LevelDB with one MongoDB for everyone. If we do that, this will all become much easier.)

ldmberman: No, we do not care about Tendermint storage here by design. After the migration, Tendermint starts from scratch while BigchainDB simply keeps the old data for querying but not for replay.

kansi:
The transaction spec for chain migration should also be included in this PR

42/README.md
```diff
@@ -70,15 +71,128 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+#### 2. Start the new chain
+
+Validators have to install and launch the new version of Tendermint. They need to prepare a new `genesis.json`. The new genesis file has to contain the validator set, the application hash (both at the migration height), and the identifier of the new Tendermint chain (`chain_id`).
+
+`chain_id` is generated and stored by BigchainDB upon conclusion of the migration election. When Tendermint sends this ID as part of the `InitChain` ABCI request, BigchainDB understands that the user has switched to a new Tendermint version so BigchainDB switches to accept new transactions and blocks.
+
+To offer a convenient way to get the data, we propose a CLI command:
+
+```
+$ bigchaindb status
```
kansi: This should be updated to `bigchaindb election status <election_id>` and the expected output (apart from `status=`) should be defined.

42/README.md
```diff
@@ -70,15 +71,124 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+With every consequent migration, the migration height is overwritten.
+
+#### Seamless HTTP API usage
+
+Since BigchainDB retains the blocks built by old Tendermint chains, the HTTP API offers the exact same experience as if there were no migrations.
+
+#### Migration election specs
+
+We introduce a new [transaction operation](./tendermint_migration_election.yaml), `TENDERMINT_MIGRATION_ELECTION`, for the purpose of implementing migration elections. `TENDERMINT_MIGRATION_ELECTION` follows the base election spec documented in [the TEP](../18).
```
kansi: `./tendermint_migration_election.yaml` is not accessible

42/README.md
```diff
@@ -70,15 +71,124 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+#### 4. Join the network after a migration
+
+When a validator joins the network after a migration, he needs to receive `genesis.json` and the chain archive directly from another member of the network or from some public place.
+
+The validator needs to restore from the archive before starting Tendermint as it is described in the previous section.
+
+Note that new validators joining a permissioned network inherently have to trust the place they are getting data from - there is no generic way to assess the validity of `genesis.json` and the archive upon joining the network.
+
+At the moment, there are no tools to verify the integrity between `genesis.json` and the archive. Such tools are subject to work on separately. In the future, one might need less trust in particular parties by downloading `genesison.json` and the archive from two different places and verifying the integrity. At the moment it is strongly recommended to download them from a single most trusted place.
```
kansi: `genesison.json`  ->  `genesis.json`

42/README.md
```diff
@@ -70,15 +71,124 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+
+- The initiator creates an election.
+- The validators vote for the election. Once the election is concluded, new blocks are not committed by the validators. New transactions from the users are rejected. New blocks sent by Tendermint are rejected too. Further in this BEP we refer to the height at which the election is concluded as the migration height. The conditions for resuming the chain operation are described below.
+
+To perform a migration election, we propose 2 CLI commands.
+
+The initiator executes:
+
+```
+$ bigchaindb election chain-migration new --private-key /home/user/.tendermint/config/priv_validator.json
```
kansi: according to the current description and implementation the command is `bigchaindb election new chain-migration ...`

42/README.md
```diff
@@ -70,15 +71,124 @@ The above process ensure the node keeps validating and accepting new blocks whil
...
+```
+
+The command outputs the migration ID. The initiator distributes it among other members of the network. The process is similar to adding new validators.
+
+To vote for an election or to see its status, validators can use common CLI described in [the BEP-18](../18).
```
kansi: How will the users get the new chain id and its validators?

PR #74, opened 2018-08-28
Made all copyright waivers more consistent with Creative Commons guidelines

ttmc:
Use some HTML markup suggested by Creative Commons, including the CC0 "Public Domain" image and a link to the Creative Commons web page with a human-readable summary of the legal code, i.e. http://creativecommons.org/publicdomain/zero/1.0/


ttmc:
@vrde 

CC0 isn't actually a license, because to license something, you have to retain the copyright. CC0 is about giving up the copyright entirely (i.e. "waiving" it).

The text was copied verbatim from the Creative Commons recommendations, and you're right, it doesn't mention CC0 explicitly. I thought that was interesting, but I'm sure they (and their lawyers) know what they're doing.

I checked, and if the image won't load, then the browser will display the text "CC0" and it will be hyperlinked. That's because of the `alt="CC0"` in the `` tag. Screen readers will also read the alt text.

PR #82, opened 2018-09-11
Extend BEP 18

ldmberman:
- Document schemas.
- Describe basic CLI.
- Add a note on the election threshold.

### DIFF
--- a/18/README.md
+++ b/18/README.md
@@ -56,6 +56,8 @@ A Validator must be able to discern valid Elections from invalid ones. A valid E
 
 **Note: any change in the Validator Set makes old Elections invalid. Check [approach 2](#generalized-approach-approach-2) for a process that can tolerate a certain degree of change to the Validator Set.**
 
+**Note 2: when final votes for multiple elections are present in the same block, all elections that do not change the Validator Set and only one election of those which do change the Validator Set are concluded.**
+
 ### Election statuses
 The lifecycle of an Election is described by the three statuses:
 
--- a/18/README.md
+++ b/18/README.md
@@ -22,11 +22,11 @@ The basic idea is to formalize the concept of an Election storing its data in a 
 ## Specification
 At any point in time, a Member of a BigchainDB Network can start a new Election.This Member is called **Initiator**.
 
-An Election is a transaction representing the matter of change, and some Vote tokens. The Initiator issues a `CREATE` transaction. The transaction has multiple `outputs`, one per Validator. Each output is populated with the public key of the Validator, and an amount equal to the power of the Validator.
+An Election is a transaction representing the matter of change, and some Vote tokens. Election transactions follow the `CREATE` transaction spec. The Initiator issues an election transaction. The transaction has multiple `outputs`, one per Validator. Each output is populated with the public key of the Validator, and an amount equal to the power of the Validator.
 
-At this point the Election starts. Independently, and asynchronously, each Validator can spend its **Vote Tokens** to an **Election Address** to show agreement on the matter of change. The Election Address is the `id` of the first `CREATE` transaction. Once the Vote tokens has been transferred to that address, it is not possible to transfer it again, because their private key is not known.
+At this point the Election starts. Independently, and asynchronously, each Validator can spend its **Vote Tokens** to an **Election Address** to show agreement on the matter of change. The Election Address is the `id` of the election transaction. Once the Vote tokens has been transferred to that address, it is not possible to transfer it again, because their private key is not known.
 
-During the `end_block` call, all transactions about to be committed are checked. Every transfer of a vote token triggers a functions that counts the number of positive votes of Election over the number of voters. If the ratio is greater than ⅔, then the current Validator commits the change. Given the BFT nature of the system, all non-Byzantine Validator will commit the change at the same block height.
+During the `end_block` call, all transactions about to be committed are checked. Every transfer of a vote token triggers a function that counts the number of positive votes of Election over the number of voters. If the ratio is greater than ⅔, then the current Validator commits the change. Given the BFT nature of the system, all non-Byzantine Validator will commit the change at the same block height.
 
 Each Validator checks every new transaction that is about to be committed in a block. The process is roughly the following:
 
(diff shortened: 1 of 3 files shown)


18/README.md
```diff
@@ -22,11 +22,11 @@ The basic idea is to formalize the concept of an Election storing its data in a
 ## Specification
 At any point in time, a Member of a BigchainDB Network can start a new Election.This Member is called **Initiator**.
 
-An Election is a transaction representing the matter of change, and some Vote tokens. The Initiator issues a `CREATE` transaction. The transaction has multiple `outputs`, one per Validator. Each output is populated with the public key of the Validator, and an amount equal to the power of the Validator.
+An Election is a transaction representing the matter of change, and some Vote tokens. Election transactions are based on `CREATE` transactions. The Initiator issues an election transaction. The transaction has multiple `outputs`, one per Validator. Each output is populated with the public key of the Validator, and an amount equal to the power of the Validator.
```
kansi: How about the following?
> An election transaction is an extension of `CREATE` transaction spec

18/README.md
```diff
@@ -22,11 +22,11 @@ The basic idea is to formalize the concept of an Election storing its data in a
...
 
-At this point the Election starts. Independently, and asynchronously, each Validator can spend its **Vote Tokens** to an **Election Address** to show agreement on the matter of change. The Election Address is the `id` of the first `CREATE` transaction. Once the Vote tokens has been transferred to that address, it is not possible to transfer it again, because their private key is not known.
+At this point the Election starts. Independently, and asynchronously, each Validator can spend its **Vote Tokens** to an **Election Address** to show agreement on the matter of change. The Election Address is the `id` of the first election transaction. Once the Vote tokens has been transferred to that address, it is not possible to transfer it again, because their private key is not known.
```
kansi: There is only one election transaction in an election DAG. I think it would suffice to state that the election address is the `id` of the election transaction.

18/README.md
```diff
@@ -42,16 +42,28 @@ A Validator must be able to discern valid Votes from invalid ones. A valid vote
...
+### Election statuses
+The lifecycle of an Election is described by the three statuses:
+
+- `ongoing`
+- `concluded`
+- `inconclusive`
+
+`ongoing` are valid Elections denoted by committed transactions, which did not receive a sufficient amount of votes, and the validator set has not changed since their creation.
+
+Elections are considered `inconclusive` if they had not been concluded by the time the validator set was changed.
```
kansi: For the sake of completeness, one line for `concluded` status could also be added to the spec.

18/README.md
```diff
@@ -92,22 +104,41 @@ If below conditions hold then the election is concluded and the proposed change
...
 
 ### Applying change
 During the `end_block` call, all transactions about to be committed are checked. Every vote token triggers a function (which implements **Approch 1** or **Approach 2**) that checks the necessary conditions. If the function returns `True` then the current Validator applies the suggested change in ![][EC_t]. Given the BFT nature of the system, all non-Byzantine Validator will commit the change at the same block height.
 
+### Command-line interface
+The interface for creating an election depends on the type of the election and has to be described by the BEP documenting this particular type.
+
+The CLI command for creating a new election should use the following template:
+```
+$ bigchaindb election <type> new ...
```
kansi: It seems that `new` keyword might not be necessary i.e. one could write `bigchaindb election upsert-validator <public_key> ...` and this should imply that the system will generate a new election

18/README.md
```diff
@@ -92,22 +106,41 @@ If below conditions hold then the election is concluded and the proposed change
...
+$ bigchaindb election show <election_id>
+```
+
+The expected output is:
+```
+status=<status>
+...
+```
+
+Additional data included in the output of the status command depends on the election type.
```
kansi: For the sake of clarity, it should be asserted that the show command must list `status=`. Other fields listed might depend on the specific election type.

18/README.md
```diff
@@ -42,22 +42,36 @@ A Validator must be able to discern valid Votes from invalid ones. A valid vote
 ### What is a valid Election?
 A Validator must be able to discern valid Elections from invalid ones. A valid Election is a transaction where **all** the following conditions are true.
 
-1. `operation` is `CREATE`, or equivalent.
-1. The `inputs` satisfy the conditions for a CREATE transaction. For example, the public key listed in `inputs.owners_before` is the public key of the Initiator.
+1. The transaction complies to the `CREATE` transaction JSON schema.
+1. The transaction additionally complies to the [election JSON schema](#./election_transaction.yaml).
```
kansi: I cannot seem to find `./election_transaction.yaml`

ldmberman: My bad, I forgot to add it.

18/election_transaction.yaml
```diff
@@ -0,0 +1,35 @@
+---
+"$schema": "http://json-schema.org/draft-04/schema#"
+type: object
+title: Election Schema
+required:
+- operation
```
kansi: `operation` is not defined in the spec

18/election_transaction.yaml
```diff
@@ -0,0 +1,35 @@
...
+- outputs
+properties:
+  asset:
```
kansi: `asset` has not been placed under `required`. Do we expect to have an election without a proposal?

18/election_transaction.yaml
ldmberman: `operation` is defined for particular election types.

18/election_transaction.yaml
ldmberman: If we do not have a reason to require it, why require?

18/election_transaction.yaml
ldmberman: Ah, yes, makes sense! I would also add a description.

18/election_transaction.yaml
kansi: Then can we remove the asset field altogether?

18/README.md
ldmberman: Discussed offline - the implementation is more or less complete and removing `new` does not seem to be worth the effort so we are keeping it.

kansi:
@ldmberman The election vote transaction spec seems missing here?

18/README.md
```diff
@@ -37,27 +37,48 @@ Each Validator checks every new transaction that is about to be committed in a b
...
+1. The transaction meets schema requirements, specific to the particular election, if any.
 1. `outputs` has as many entries as the total number of Validators.
 1. Each Validator is represented in `outputs`.
 1. Each entry in `outputs` can be spent by only one Validator, and the amount attached to it is equal to the power of that Validator.
 
-**Note: any change in the Validator Set will make old Elections invalid. Check [approach 2](#generalized-approach-approach-2) for a process that can tolerate a certain degree of change to the Validator Set.**
+**Note: any change in the Validator Set makes old Elections invalid. Check [approach 2](#generalized-approach-approach-2) for a process that can tolerate a certain degree of change to the Validator Set.**
+
+**Note 2: when final votes for multiple elections are present in the same block, all elections that do not change the Validator Set and only one election of those which do change the Validator Set are concluded.**
```
vrde: I don't understand why we need this note. If there are multiple final votes, they will be executed using the order of the transactions in the block, otherwise we might end up in adding logic to sort elections based on some rules.

18/README.md
vrde: most of the commands are `bigchaindb election approve/show <election_id>`, while this one is `bigchaindb election <type> new`, would it make sense to have `bigchaindb election new <type>`?

Like in `git`, you have `git remote add/remove/show <...>`

18/README.md
ldmberman: It explains the extension of this rule which allows elections to be concluded even though one of them changes the validator set. Votes are still processed in their original order but we differentiate between elections that change the validator set and those which do not.

We found it convenient and easy to do but I am also fine with dropping this rule and condemning all the elections that follow the one which changes the validator set to be inconclusive, even within the same block.

ldmberman:
> would it make sense to have bigchaindb election new <type>?

This is actually how it was implemented, I'll fix it here.

18/README.md
vrde: In my opinion, the simpler we can keep the election protocol the better, for me you can drop this rule.

ldmberman: Yeah, I would also prefer the simplicity actually - this way how we handle multiple elections within the same block would be consistent with how we do it across blocks.

PR #85, opened 2018-09-14
Problem: there is no performance analysis of bdb

vrde:
Solution: run some performance tests and write a report about it.


ttmc:
Did you write the transaction size (or average transaction size) in the report? I know that information was captured in the CSV files. We need to share that number.

23/README.md
```diff
@@ -0,0 +1,280 @@
...
+type: Informational
+status: Raw
+editor: Alberto Granzotto <alberto@bigchaindb.com>
+contributors: Troy McConaghy <troy@bigchaindb.com>, Muawia Khan <muawia@bigchaindb.com>
+```
+
+# Performance Study: Analysis of Transaction Throughput in a BigchainDB Network
+
+## Abstract
+Transaction throughput is a key metric to evaluate the suitability of a blockchain system for real world applications. BigchainDB is now in its beta phase, and we don't expect major changes in the parts of the code that are responsible for the throughput of the system (i.e. the validation logic), for this reason we can now run performance tests and publish our results. The tested setup consists of a 4 node network running a [recent version of BigchainDB][bdb:git-commit]. Each node runs on a different virtual machine, hosted in the [Azure infrastructure][azure:landing-page]. A fifth virtual machine is used to generate the workload and collect metrics. Depending on the configuration of the system, a BigchainDB network under constant load finalizes between 300 and 1000 transactions per second.
```
ttmc: The last sentence of this abstract is the "money quote" so we need to make sure it's absolutely accurate. I'd change it slightly, to say:

Depending on the configuration of the system, the tested BigchainDB network, under constant load, finalizes between 300 and 1000 transactions per second.

vrde:
> Did you write the transaction size (or average transaction size) in the report? I know that transaction size information was captured in the CSV files. We need to share that number.

Good point, I'll add it right now.

23/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+Those metrics reflect the different states of a valid transaction: accepted and finalized.
+
+Even if the reference metric is how many transactions the network finalizes per second, during this analysis we dig deeper and analyze the distribution on accepted transactions, committed etc. We want to make sure that the system is responsive most of the time, and we don't want to get fooled by an average that hides a lot of data behind it. Knowing the actual distribution of how much time it takes for a transaction to be accepted and then committed allows us to have a clear understanding of the responsiveness of the system, and this is something we won't know if we look only at the average. For this reason we developed our own tool, `bigchaindb-benchmark`.
+[BigchainDB Benchmark][bdb:benchmark-tool] is the tool we developed to generate the workload and collect statistics. It can simulate a large amount of workload by spawning multiple processes, and output a CSV file to timestamp the different steps in the life cycle of a transaction. This allows us to measure the distribution of how long it takes for a transaction to be accepted and later finalized.
```
ttmc: These sentences are repetitive, i.e.

> For this reason we developed our own tool, `bigchaindb-benchmark`. [BigchainDB Benchmark][bdb:benchmark-tool] is the tool we developed to generate the workload and collect statistics.

They could be simplified into one sentence such as:

> For this reason we developed our own benchmarking tool to generate the workload and collect statistics; we call it [bigchaindb-benchmark][bdb:benchmark-tool].

23/README.md
```diff
@@ -0,0 +1,283 @@
...
...block, it can parallelize validation, exploiting all cores in the current machine (more details to come). To understand if we could get a performance boost out of this, we integrated a new experimental feature that enables parallel validation of transactions. This feature is a proof of concept to show that higher throughput is realizable. Note: this feature skips also *Check Tx*, used by Tendermint to check a transaction before storing it in the mempool. BigchainDB validates transactions as soon as they are posted to the HTTP API, so *Check Tx* in this case is redundant. We must mention that *Check Tx* is used also to validate transactions coming from the other nodes of the network, so disabling it might allow byzantine actors to push invalid transactions to the mempool (handling invalid transactions is [actively discussed in the Tendermint issue tracker][tm:invalid-txs]). Even if an invalid transaction makes it to the mempool, it will be always discarded during the *Deliver Tx* phase, so this experimental feature will still maintain a correct state of the application. Parallel validation is an optimistic approach that will be eventually developed for specific production use cases.
```
ttmc: `Note: this feature skips also *Check Tx*,`

should say

`Note: this feature also skips *Check Tx*,`

and I wonder if we should write it CheckTx like how Tendermint does.

23/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Methods
+<!--
+METHODS. The methods section will help you determine exactly how the authors performed the experiment.
+
+The methods describes both specific techniques and the overall experimental strategy used by the scientists. Generally, the methods section does not need to be read in detail. Refer to this section if you have a specific question about the experimental design.
+-->
+
+The performance tests were run on the Microsoft Azure cloud infrastructure. For the BigchainDB nodes, we choose compute optimized virtual machines from the [Fsv2 series][azure:fsv2], specifically *Standard F32s v2* (32 vCPUs, 64GiB memory, SSD drives with estimated IOPS limit of 5000). For the machine generating the workload, we used a *Standard F8s v2* (8 vCPUs, 16GiB memory). The virtual machines were all in the same datacenter.
```
ttmc: I checked the linked chart about the Fsv2 series VMs and it says that for Standard_F32s_v2 VMs:

Max cached and temp storage throughput: IOPS / MBps (cache size in GiB) is 64000 (512)

and

Max NICs / Expected network bandwidth (Mbps) is 8 / 14,000

23/README.md
```diff
@@ -0,0 +1,283 @@
...
+benchmark1@benchmark1:~$ sudo hdparm -Tt /dev/sda
+
+/dev/sda:
+ Timing cached reads:   19514 MB in  1.99 seconds = 9785.28 MB/sec
+ Timing buffered disk reads: 1484 MB in  3.00 seconds = 494.39 MB/sec
+```
+
+All machines were provided with the operating system *Ubuntu 18.04.1 LTS (Bionic Beaver)*. For simplicity we provided all machines (both the machine generating the workload and the machines running the BigchainDB nodes) with the same software:
+
+- BigchainDB, [commit 6a9064196ad49baf3933051be6e116a3440fbad2][bdb:git-commit]
```
ttmc: Giving the specific Git commit is helpful for coders and anyone trying to replicate the results, but it's not helpful for the average reader. Maybe add something to give them a sense of where that commit fits into the BigchainDB version history, like 

> BigchainDB, commit blahblahblah (between version 2.0.0b5 and 2.0.0b6)

23/README.md
```diff
@@ -0,0 +1,283 @@
...
+- MongoDB v3.6.3
+
+This procedure was automated by a [custom script][self:provide.sh].
+
+Three main configuration files were generated and edited (all of them are included as extra files in this BEP):
+- The Tendermint [`config.toml`](self:config.toml) configuration file (default location: `${HOME}/.tendermint/config/config.toml`) was generated via `tendermint init` and customized.
+- The Tendermint [`genesis.json`][self:genesis.json] file (generated though the previous command `tendermint init`) was customized with the public keys of the four nodes. It was then copied into all machines under `${HOME}/.tendermint/config/genesis.json`.
+- The BigchainDB [`.bigchaindb`][self:bigchaindb.json] was generated via `bigchaindb configure` and copied into all machines under `${HOME}/.bigchaindb`.
+
+We also had a **setup** and **teardown** procedure we run before and after every test.
```
ttmc: `run` --> `ran`

23/README.md
```diff
@@ -0,0 +1,283 @@
...
+For each node in the network, we started all services using the startup script. After some seconds, we switched to the workload-generating virtual machine, and ran:
+
+```
+bigchaindb-benchmark\
+    --csv results/master-32proc-1000000sync.csv\
+    --processes=32\
+    --peer http://10.240.0.4:9984\
+    --peer http://10.240.0.5:9984\
+    --peer http://10.240.0.6:9984\
+    --peer http://10.240.0.7:9984\
```
ttmc: Should we comment that if they try to repeat our experiments, then their IP addresses will almost certainly be different?

23/README.md
```diff
@@ -0,0 +1,283 @@
...
+    send -r1000000 --mode=sync
+```
+
+This command uses 32 workers to push 1,000,000 transaction of 765 bytes to the four nodes in the network. The generated `CSV` file with the results was then copied to another machine
```
ttmc: The second sentence here is missing a period at the end (after "machine"). Same for similar sentences below.

23/README.md
vrde: About the IOPS of the disk, I used the data coming from portal.azure.com... Not sure why they have two different numbers, check this screenshot:
image


I'll make the suggested changes.

23/README.md
vrde: Yes, I'm reporting what we used. Don't want to sound rude, but if they don't understand this, I'm not sure they can actually do a benchmark. But if we want to add a comment about that, we should write it at the beginning of the document, because it applies also to the configuration files presented in this BEP.

23/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+We built a script called [`analysis.py`][bdb:analysis.py] to process those CSV files and output four different graphs per experiment:
+- Distribution of how much time it takes to **accept** a transaction.
+- Distribution of how much time it takes to **finalize** a transaction.
+- **Accepted** transactions per second.
+- **Finalized** transactions per second.
+
+
+### Experiment 1: finalize 1,000,000 transactions (master)
+![Experiment 1][self:plot-experiment-1]
```
ttmc: GitHub renders the SVG file quite small, and clicking on it doesn't help. There doesn't seem to be any easy way to zoom in, even though in theory it is a vector graphic and therefore infinitely zoom-in-able. Maybe we could just embed a big JPEG for GitHub readers and provide the SVG download link if people want that.

23/README.md
ttmc: Let's add a comment _somewhere_ about how, if they want to reproduce our results, their node IP addresses will be different.

23/README.md
```diff
@@ -0,0 +1,283 @@
...
+```
+BigchainDB was able to process 1,000,000 transactions in 26 minutes without any failure.
+
+In terms of responsiveness, the system performed consistently during the whole test:
+- 99.7% of transactions have been **accepted** within 2.033 seconds, 95% within 0.113 seconds, and 68% within 0.018 seconds.
+- 99.7% of transactions have been **finalized** within 7.686 seconds, 95% within 5.402 seconds, and 68% within 2.898 seconds.
+
+In terms of speed (transactions per second), the system:
+- **Accepted** an average of _636 transactions per second_, with a median value of _615 transactions per second_.
+- **Finalized** an average of _636 transactions per second_, with a median value of _636 transactions per second_.
```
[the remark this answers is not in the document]
vrde: The test run for 26.2 minutes, I dropped the .2. I can write the number of seconds (1572) and say that is about 26 minutes, what do you think?

23/README.md
vrde: What about adding at the beginning of the **Methods** section the following sentence:
> Note: to reproduce our results, make sure to adapt the configuration files, scripts, and commands to your setup.

23/README.md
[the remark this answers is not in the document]
ttmc: Saying "about 26" minutes might be better, yes.

23/README.md
ttmc: That sounds quite open-ended and scary. Maybe you could make it less scary by writing a specific example of something they will have to change:

> Note: to reproduce our results, make sure to adapt the configuration files, scripts, and commands to your setup. For example, your nodes will have different IP addresses.

23/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Discussion
+<!--DISCUSSION. The discussion section will explain the authors interpret their data and how they connect it to other work.
+
+Authors often use the discussion to describe what their work suggests and how it relates to other studies. In this section, authors can anticipate and address any possible objections to their work. The discussion section is also a place where authors can suggest areas of improvement for future research.-->
+
+
+The aims of the study are twofold, on one side we want to have an understanding on how a BigchainDB network performs under stress; on the other side we want to define a methodology we can use for a) future versions of BigchainDB, b) running other tests under different conditions.
+
+About the performance of the system, we can conclude that BigchainDB performs consistently and correctly under constant stress, and it's able to finalize one million transactions within one hour and always be responsive. Enabling parallel validation allows BigchainDB to finalize the same number of transactions less than half an hour.
```
ttmc: `finalize the same number of transactions less than half an hour.`

should be

`finalize the same number of transactions in less than half an hour.`

23/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+We also ran a short test to measure the peak performance of the system: under favorable conditions we are able to reach ~1000 finalized transactions per second. We are confident that we can improve the system and reach this performance under constant, heavy load.
```
ttmc: Let's remove that last sentence in case it comes back to bite us.

23/README.md
vrde: It sounds much better! :smile_cat:

23/README.md
vrde: I've improved the SVG files, take a look and tell me what you think.

23/README.md
vrde: :snake:

23/README.md
ttmc: I checked and the SVG files _do_ look much better now. I can now read the labels. Issue resolved. Disaster averted.

PR #89, opened 2018-09-17
Updated state of BEP-15

gautamdhameja:


ttmc:
We can add a link to the implementation once that's public. I'll merge this now.

